PR #3, opened 2023-06-05
Setup a basic Netty MQTT server

antonio-pedro99:
In this PR:

1. MqttServer: 

- Class responsible for starting the Socket Channel using ChannelInitializer from Netty framework:
          - Defines the MqttDecoder, Encoder and relies on custom MqttServerHandler to handle the communication.
 - Two methods: 
          - start(): starts the Future channel, uses the server bootstrap to bind the port and sync.
          - stop(): shutdown the server gracefully.
2. MqttServerHandler:
- Inherits the SimpleChannelInboundHandler from Netty framework.
- Listens to active client connections(channel)
- Listens to incoming messages and check if they are MQTT type messages:
      - If message is CONNECT, then send an ConnAckMessage to the client.
      - if message is PUBLISH, tells MqttKafkaMapper to start the mapping of the incoming MQTT message.
 3. MqttKafkaMapper:
 - Concurrently, try to perform the mapping operation using a combination of threads and runnable.
 - Right now it is just printing whatever the message and the topic are to the console.
 
Screenshots:
I am running the server normally on my local machine
image

For testing, I am using MQTT Explorer to connect to the server.

image
Once the client is connected 

image

We are getting two SUBSCRIBE messages because the MQTT explorer try to subscribe to default topics($SYS/broker/clients/connected and  $SYS/broker/messages/sent topics) to retrieve info about the broker(yes, it assumes that it just connect to brokers).

Clients can now publish messages to the server
image
Result
image

I also tested the server using my phone(through my local network).

**How to Test**
download https://gist.github.com/antonio-pedro99/1c7c2132aa25581f37e16249b00a62e6
install paho
`pip install paho-mqtt` <br>

Run `python3 test_script.py`


antonio-pedro99:
@ppatierno @kyguy

mqtt-bridge/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+        <maven.compiler.source>19</maven.compiler.source>
+        <maven.compiler.target>19</maven.compiler.target>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport</artifactId>
+            <version>4.1.87.Final</version>
```
ppatierno, line 21: why using a different version from the other Netty dependencies?
Also it's important to use a property to define the Netty version in the `<properties>` section so that it can be reused easily across different Netty dependencies.

mqtt-bridge/src/main/java/io/strimiz/strimzi_mqtt_bridge/Main.java
```diff
@@ -0,0 +1,20 @@
+package io.strimiz.strimzi_mqtt_bridge;
```
ppatierno, line 1: typo in the first part of the package `io.strimiz`.
I would also use the same pattern we have for the Strimzi HTTP bridge which is `io.strimzi.kafka.bridge.http` so in this case by using `io.strimzi.kafka.bridge.mqtt`. Using underscores in the package names is a bad pattern.

mqtt-bridge/src/main/java/io/strimiz/strimzi_mqtt_bridge/core/MqttServerHandler.java
```diff
@@ -0,0 +1,57 @@
...
+
+
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.SimpleChannelInboundHandler;
+import io.netty.handler.codec.mqtt.*;
+
+public class MqttServerHandler extends SimpleChannelInboundHandler<Object> {
+    @Override
+    public void channelActive(ChannelHandlerContext ctx) {
+        System.out.printf("Client " + ctx.channel().remoteAddress() +  " is trying to connect\n");
```
ppatierno, line 11: need for a logging library and not using `System.out`. We should go with using log4j2 in all classes.

mqtt-bridge/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>io.strimzi.strimzi_mqtt_bridge</groupId>
+    <artifactId>strimzi_mqtt_bridge</artifactId>
+    <version>1.0-SNAPSHOT</version>
+
+    <properties>
+        <maven.compiler.source>19</maven.compiler.source>
+        <maven.compiler.target>19</maven.compiler.target>
```
ppatierno, line 13: let's use `17` for both, as all the Strimzi project are using Java 17 compiler for now.

mqtt-bridge/pom.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>io.strimzi.strimzi_mqtt_bridge</groupId>
```
ppatierno, line 7: should be just `io.strimzi`

mqtt-bridge/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+    <artifactId>strimzi_mqtt_bridge</artifactId>
```
ppatierno, line 8: better just `mqtt-bridge`

mqtt-bridge/src/main/java/io/strimiz/strimzi_mqtt_bridge/Main.java
```diff
@@ -0,0 +1,20 @@
...
+
+import io.netty.channel.ChannelOption;
+import io.netty.channel.nio.NioEventLoopGroup;
+import io.strimiz.strimzi_mqtt_bridge.core.MqttServer;
+
+public class Main {
+    static final int PORT = Integer.parseInt(System.getProperty("port", "1883"));
```
ppatierno, line 8: You should start to think about a more complete mechanism to provide full configuration.
I would assume that the MQTT `port` is just one of them but at some point we would also allow the users to configure the Kafka producer part of the bridge.
See for example how it's done today in the Strimzi HTTP bridge as a reference.
Maybe not something to add to this PR of course.

ppatierno (changes requested):
I had a very quick first pass but I need to review the logic.
I left some comments and also:
- you should not push `.idea` folder in your PRs. This is IDE specific not part of the project.
- you should not push `target` folder containing the `.class` files.
- you should add Javadoc for each classes and methods. Let's start from the beginning otherwise you will end with a lot of Javadoc to be added and maybe even not remembering the purpose of a method (at least it happens to me ;-)).

mqtt-bridge/src/main/java/io/strimiz/strimzi_mqtt_bridge/core/MqttServerHandler.java
```diff
@@ -0,0 +1,57 @@
+package io.strimiz.strimzi_mqtt_bridge.core;
+
+
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.SimpleChannelInboundHandler;
+import io.netty.handler.codec.mqtt.*;
```
ppatierno, line 6: As code style in Strimzi we avoid using `*` for importing classes. It's better to list what's used here.

ppatierno (changes requested):

mqtt-bridge/pom.xml
antonio-pedro99: Actually it was not intentional, I was using the IDE to automatically and the dependencies, I think it was not picking the same versions. 
but I have fixed it now.

mqtt-bridge/src/main/java/io/strimiz/strimzi_mqtt_bridge/Main.java
antonio-pedro99: Noted sir. I have made the required changes.

mqtt-bridge/src/main/java/io/strimiz/strimzi_mqtt_bridge/core/MqttServerHandler.java
antonio-pedro99: Noted sir, in this case I just used sout for rapid testing.
I am working on it.

mqtt-bridge/src/main/java/io/strimiz/strimzi_mqtt_bridge/Main.java
antonio-pedro99: Yes, I am keeping that in mind.
The same thing holds for the javadoc in classes.

.idea/strimzi-mqtt-bridge.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="JAVA_MODULE" version="4">
+  <component name="NewModuleRootManager" inherit-compiler-output="true">
+    <exclude-output />
+    <content url="file://$MODULE_DIR$" />
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
```
kyguy, line 9: This file should be removed as well?

mqtt-bridge/.gitignore
```diff
@@ -0,0 +1,17 @@
...
+.apt_generated/
+.apt_generated_tests/
+
+# IntelliJ IDEA specific
+.idea/
+*.iml
+
+# VS Code
+.factorypath
+.vscode
```
kyguy, line 17: I would copy the `.gitignore` file of the strimzi-kafka-bridge (HTTP bridge) [1]

And place this file in the parent directory


[1] https://github.com/strimzi/strimzi-kafka-bridge/blob/main/.gitignore

mqtt-bridge/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+            <version>${netty.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-all</artifactId>
+            <version>${netty.version}</version>
+        </dependency>
+    </dependencies>
+
+</project>
```
kyguy, line 36: Since this repo is dedicated to the Strimzi MQTT bridge, should we move files like these to the parent directory and remove the mqtt-bridge directory?

kyguy (commented):
Hey Antonio I left a few minors comments concerning the format of the repo, but overall an awesome start! 

Could you add some instructions on how to build and run the code you have so far? The HTTP bridge has something like that here [1], maybe we start with some simple instructions and evolve to follow that format to be consistent.

[1] https://github.com/strimzi/strimzi-kafka-bridge/blob/main/BUILDING.md

mqtt-bridge/.gitignore
antonio-pedro99: done as you requested.

.idea/strimzi-mqtt-bridge.iml
antonio-pedro99: Regarding [1], I will look into it further. But do we need the building in this PR?

ppatierno: `.iml` is IntelliJ specific we don't want this stuff to be pushed. People can use any other IDE (i.e. VSCode) for developing.

mqtt-bridge/pom.xml
ppatierno: +1

ppatierno: @antonio-pedro99 it's not just about the pom.xml, I didn't notice that everything is under an `mqtt-bridge` folder which should not exist. Everything has to be under the root, just move the content of `mqtt-bridge` folder in the root and you should be done.

mqtt-bridge/src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -0,0 +1,82 @@
...
+import io.netty.handler.codec.mqtt.MqttConnAckMessage;
+
+/**
+ * Represents a SimpleChannelInboundHandler. The MqttServerHandler is responsible for: <br>
+ * - listen to client connections;<br>
+ * - listen to incoming messages; <br>
+ *
+ * @see io.netty.channel.SimpleChannelInboundHandler
+ */
+public class MqttServerHandler extends SimpleChannelInboundHandler<Object> {
```
ppatierno, line 20: I guess you can use `MqttMessage` directly (as concrete type) and not casting the `Object` msg later in the `channelRead`, so something like ...

```java
public class MqttServerHandler extends SimpleChannelInboundHandler<MqttMessage>
```

mqtt-bridge/src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -0,0 +1,82 @@
...
+    @Override
+    public void channelActive(ChannelHandlerContext ctx) {
+        System.out.printf("Client " + ctx.channel().remoteAddress() +  " is trying to connect\n");
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) {
+        try {
+            MqttMessageType messageType = ((MqttMessage) msg).fixedHeader().messageType();
```
ppatierno, line 29: as per the suggestion above you could just have ...

```java
@Override
public void channelRead(ChannelHandlerContext ctx, MqttMessage msg) {
     try {
            MqttMessageType messageType = msg.fixedHeader().messageType();
```

ppatierno (changes requested):

ppatierno:
@antonio-pedro99 I see you requested review again (which is anyway not needed to do every time because we get notifications from GitHub when you change something) but anyway I still see some comments not addressed yet.

antonio-pedro99:
till now, I think only the Logging and project folder refactor is missing, am I wrong?

ppatierno:
@antonio-pedro99 yes ... well more or less ... I have something I am thinking more about and will update the PR soon.
Anyway, I think that logging and removing the mqtt-bridge folder are must have for this PR.
About logging it should be pretty easy, just adding a log4j2 and slf4j dependencies, a log4j2.properties configuration (in resources) and using it :-) Take a look at what you need in the Strimzi HTTP Kafka bridge.

antonio-pedro99:
Yes, I was already doing that xD, somehow it was not logging anything so I left it for a while. I am working on it now.

antonio-pedro99:
Sure! Meanwhile, all the changes requested have been solved.

This was supposed to be the system design improvement :)

antonio-pedro99:
@ppatierno Regarding the Mapper using threads.
The mapper will be doing heavy work → mapping, which might take some computation time. Something I am sure is that this mapping should be done concurrently.
I used threads because it was the easiest thing I could think about for this PR, as I said, I was just playing with Netty during the weekend :). But I am happy to take a look further a look for better alternatives, maybe using the Netty utilities for concurrency and async tasks is a good start. 

Regarding it being a singleton, I think it is trivial, and we do not need to create multiple instance of it. By trivial, I mean, creating an instance of the mapper inside the MqttServerHandler.java would do the work.

ppatierno:
I think the question is how the Netty thread model works. It needs further investigation. It depends how each channel handler is executed in its own thread or not and if Netty has a thread pool. In that case we could not need a new thread doing all the work for requests coming from channel handlers (for multiple connections). One thing you could do is about logging the message in the `handlePublishMessage` directly (no mapper usage) and print the Thread name as well. Then writing a script (bash? python?) which open multiple MQTT connections and publish messages and see how the server reacts and how many threads are used by Netty.

antonio-pedro99:
@ppatierno  sure. Investigating how netty threads works is what exactly I am doing now. 
I am following the procedure suggested by you and might get you back in few hours.

antonio-pedro99:
I have edited the description of this PR with the instructions to test the server using a python script.
1. execute the server
2. open the terminal in the root folder
3. run python3 test_script.py

ppatierno:
@antonio-pedro99 we need a license header at the top of all the `.java` files. Would you mind to add the following one on all of them, please?

```shell
/*
 * Copyright Strimzi authors.
 * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
 */
```

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServer.java
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * Start the server.
+     */
+    public void start() {
+        try {
+            ChannelFuture channelFuture = this.serverBootstrap.bind(this.port).sync();
+            channelFuture.channel().closeFuture().sync();
+        } catch (InterruptedException e) {
+            logger.error(e.getMessage());
```
ppatierno, line 70: logger.error("Error starting the MQTT server: ", e);

src/main/java/io/strimzi/kafka/bridge/mqtt/Main.java
```diff
@@ -0,0 +1,23 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(Main.class);
+    static final int PORT = Integer.parseInt(System.getProperty("port", "1883"));
+
+    public static void main(String[] args) {
+        try {
+            NioEventLoopGroup masterGroup = new NioEventLoopGroup();
+            NioEventLoopGroup workerGroup = new NioEventLoopGroup();
+            MqttServer server = new MqttServer(PORT, masterGroup, workerGroup, ChannelOption.SO_KEEPALIVE);
+            server.start();
+        } catch (Exception e) {
```
ppatierno, line 19: which exception this is supposed to catch? You are calling just `start` on the mqtt server and it's already catching exception and stopping the server if needed.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -0,0 +1,84 @@
...
+            if (messageType == MqttMessageType.CONNECT) {
+                handleConnectMessage(ctx);
+            } else if (messageType == MqttMessageType.PUBLISH) {
+                MqttPublishMessage message = (MqttPublishMessage) msg;
+                message.retain();
+                handlePublishMessage(ctx, message);
+                message.release();
+            } else {
+                // Handle other MQTT message types as needed
+                logger.debug("Got {} message type", messageType.name());
```
ppatierno, line 44: this should be at the top right after checking the message type, we would log all messages. In this case you are logging everything but CONNECT and PUBLISH.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -0,0 +1,84 @@
...
+            }
+        } catch (Exception e) {
+            logger.error(e.getMessage());
```
ppatierno, line 47: See my previous comment about logging with a text and providing the Exception instance directly

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * Handle the case when a client sent a MQTT PUBLISH message type.
+     *
+     * @param ctx            ChannelHandlerContext instance
+     * @param publishMessage represents a MqttPublishMessage
+     */
+    private void handlePublishMessage(ChannelHandlerContext ctx, MqttPublishMessage publishMessage) {
+        logger.info("MAPPING...");
```
ppatierno, line 80: I would delete this log imho.

ppatierno (changes requested):

ppatierno (approved):
LGTM. Thanks for the PR!

kyguy (approved):
Looks great @antonio-pedro99, compiles, runs, and behaves just as expected, nice work!

antonio-pedro99:
@ppatierno @kyguy thanks 😊

PR #5, opened 2023-06-14
add: added the Mapper component

antonio-pedro99:


src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,66 @@
...
+        rules = new ArrayList<>();
+        rules.add(new MappingRule("building_{building}_room_{room}", "building/{building}/room/{room}/#", null));
+        rules.add(new MappingRule("sensor_data", "sensors/+/data", null));
+        rules.add(new MappingRule("devices_{device}_data", "devices/{device}/data", null));
+        rules.add(new MappingRule("fleet_{vehicle}", "fleet/{fleet}/vehicle/{vehicle}/#", null));
+        rules.add(new MappingRule("building_{building}_others","building/{building}/#", null));
+        rules.add(new MappingRule("sensor_others", "sensors/#", null));
+        rules.add(new MappingRule("building_others", "building/#", null));
+        rules.sort(Comparator.comparing(MappingRule::getMqttTopicPatternLevels).reversed());
+    }
```
ppatierno, line 37: Why does the constructor fill all these hard coded rules? They should be loaded from a JSON file as we agreed or anyway if not loading from a file yet, you should be able to fill different rules during different tests to check that the logic works.

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,66 @@
...
+     *
+     * @param mqttTopic represents an MQTT topic.
+     * @return a valid Kafka topic.
+     * @see MappingRule
+     */
+    public String map(String mqttTopic) {
+        for (MappingRule rule : this.rules) {
+            //convert the mqtt pattern to a valid regex expression.
+            String regex = rule.getMqttTopicPattern().replaceAll("\\{\\w+\\}", "(.*)")
+                    .replace("#", ".*").replace("+", "[^/]+");
```
kyguy, line 50: It might be good to: 
(1) Abstract these into final variable fields for the class with a comment describing in English what the regex pattern matches. 
(2) Add comments above the lines that call replace() with an example of what is happening

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -78,7 +79,7 @@ private void handleConnectMessage(ChannelHandlerContext ctx) {
      * @param publishMessage represents a MqttPublishMessage
      */
     private void handlePublishMessage(ChannelHandlerContext ctx, MqttPublishMessage publishMessage) {
-        logger.info("Topic: {}", publishMessage.variableHeader().topicName());
-        logger.info("Message: {}", publishMessage.payload().toString(Charset.defaultCharset()));
+        logger.info("MQTT Topic: {}", publishMessage.variableHeader().topicName());
+        logger.info("Mapped Topic: {}", mapper.map(publishMessage.variableHeader().topicName()));
```
kyguy, line 83: Is the "Mapped Topic" the name of the Kafka topic?

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRule.java
```diff
@@ -0,0 +1,65 @@
...
+    private HashMap<String, String> placeholdersType = new HashMap<>();
+
+    /**
+     * Constructor
+     *
+     * Creates a new instance of MappingRule.
+     * @param kafkaTopicTemplate the desired Kafka topic template. The kafka topic template can contain placeholders.
+     * @param mqttTopicPattern the mqtt topic pattern. The mqtt topic pattern can contain placeholders.
+     */
+    public MappingRule (String kafkaTopicTemplate, String mqttTopicPattern, HashMap<String, String> placeholdersType){
```
kyguy, line 25: What is the placeholdersType parameter? (Don't forget to add the @param Javadoc here too!)

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRule.java
```diff
@@ -0,0 +1,65 @@
...
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.bridge.mqtt.utils;
+
+import java.util.HashMap;
+
+/**
+ * Represents a Mapping Rule in the Topic Mapping Rules(TOMAR).
+ *
+ */
```
kyguy, line 12: Might be good to include an example in the class comment up here of what a user supplied MappingRule would look like

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
```diff
@@ -0,0 +1,68 @@
...
+                map("building/101"), is("building_others"));
+    }
+
+    /**
+     * Help method to map an MQTT topic to a Kafka topic.
+     */
+    protected String map(String mqttTopic){
+        return mqttKafkaMapper.map(mqttTopic);
+    }
+}
```
kyguy, line 68: The MappingRule that is used as part of these tests (e.g. the ones hardcoded in MqttKafkaMapper() ) should be defined in this test class or loaded in via a separate isolated JSON file (as Paolo mentioned) dedicated for testing

kyguy (commented):
Great start @antonio-pedro99, left some comments but keep up the momentum

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: loading during the tests should be fine. But I am already working on loading configurations from application.proprieties file.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
antonio-pedro99: the Mapped Topic is the result of the mapping, yes, it is the Kafka topic.

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRule.java
antonio-pedro99: it would hold data type for each placeholder, e.g device/{id}/data, we should be able to define which data type the {id} placeholder will be of. 
but I am removing it as we do not need that now.

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRule.java
antonio-pedro99: @kyguy I agree.

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
antonio-pedro99: I agree. I am working on it.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: I agree, it would make the code more readable.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: for this PR, I am loading the test cases from a json.

ppatierno: I am not sure that for testing this we need a JSON file with a lot of rules to be loaded.
I think we should have more different tiny tests focused on some rules that you can pass to the `MqttKafkaMapper` easily. In the end, it's a list of `MappingRule`. So each test should fill a rules list with a few rules and then testing just them.

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
```diff
@@ -0,0 +1,103 @@
...
+    //define the mapping rules as a JSON string. Add more rules if needed.
+    private static final String TOPIC_MAPPING_RULES_JSON =
+            "[" +
+                    "{\"kafkaTopic\":\"building_{building}_room_{room}\",\"mqttTopic\":\"building/{building}/room/{room}/#\"}," +
+                    "{\"kafkaTopic\":\"sensor_data\",\"mqttTopic\":\"sensors/+/data\"}," +
+                    "{\"kafkaTopic\":\"devices_{device}_data\",\"mqttTopic\":\"devices/{device}/data\"}," +
+                    "{\"kafkaTopic\":\"fleet_{vehicle}\",\"mqttTopic\":\"fleet/{fleet}/vehicle/{vehicle}/#\"}," +
+                    "{\"kafkaTopic\":\"building_{building}_others\",\"mqttTopic\":\"building/{building}/#\"}," +
+                    "{\"kafkaTopic\":\"sensor_others\",\"mqttTopic\":\"sensors/#\"}," +
+                    "{\"kafkaTopic\":\"building_others\",\"mqttTopic\":\"building/#\"}" + "]";
```
kyguy, line 36: Could you format this String in the same manner the standalone JSON is formatted? IDC how many more lines it takes, I think it would be easier to read

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
```diff
@@ -0,0 +1,103 @@
...
+                map("sensors/25/temperature/data"), is("sensor_others"));
+
+        assertThat("Mqtt pattern sensors/# should be mapped to sensor_others",
+                map("sensors/25/data/humidity"), is("sensor_others"));
+
+        assertThat("Mqtt pattern devices/{device}/data should be mapped to devices_{device}_data",
+                map("devices/4/data"), is("devices_4_data"));
+
+        assertThat("Should use the default topic when no mapping rule is found",
+                map("sensor/temperature"), is("messages_default"));
```
kyguy, line 72:
```suggestion
                map("sensor/temperature"), is(DEFAULT_KAFKA_TOPIC));
```

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Load all the mapping rules before running the tests.
+     */
+    @Before
+    public void setUp() {
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            mappingRules = mapper.readValue(TOPIC_MAPPING_RULES_JSON, new TypeReference<>() {
+            });
```
kyguy, line 46: Any reason this isn't with the line above?

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
kyguy: Could we call it Kafka Topic instead?
```suggestion
        logger.info("Kafka Topic: {}", mapper.map(publishMessage.variableHeader().topicName()));
```

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRule.java
```diff
@@ -0,0 +1,64 @@
...
+ * {
+ * "kafkaTopic": "sensors_{sensorId}_data",
+ * "mqttTopic": "sensors/{sensorId}/data"
+ * }
+ * and like this in the MappingRule class:
+ * MappingRule(kafkaTopicTemplate=sensors_{sensorId}_data mqttTopicPattern= sensors/{sensorId}/data)
+ */
+public class MappingRule {
+    @JsonProperty("kafkaTopic")
+    private String kafkaTopicTemplate;
```
kyguy, line 21: Would it make more sense to call this `kafkaTopicPattern` instead of `kafkaTopicTemplate` for symmetry with `mqttTopicPattern`?

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,68 @@
...
+     */
+    public String map(String mqttTopic) {
+        for (MappingRule rule : this.rules) {
+            //convert the mqtt pattern to a valid regex expression.
+            // the mqtt pattern can contain placeholders like {something}, + and #.
+            // if the mqtt topic contains a +, we replace it with @singleLevelWildcardRegex
+            // if the mqtt topic contains a #, we replace it with @multiLevelWildcardRegex
+            // if the mqtt topic contains a placeholder (pattern \{\w+\}), we replace it with @placeholderRegex
+            String regex = rule.getMqttTopicPattern().replaceAll("\\{\\w+\\}", PLACEHOLDER_REGEX)
+                    .replace("#", MULTIPLE_LEVEL_WILDCARD_REGEX).replace("+", SINGLE_LEVEL_WILDCARD_REGEX);
```
kyguy, line 52: Would it make sense to translate the MappingRules to regex outside of the method so we only have to do it once instead of every time we have to map a mqttTopic name? Maybe we could do it as part of the MappingRules class?

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,68 @@
...
+     * @see MappingRule
+     */
+    public String map(String mqttTopic) {
+        for (MappingRule rule : this.rules) {
+            //convert the mqtt pattern to a valid regex expression.
+            // the mqtt pattern can contain placeholders like {something}, + and #.
+            // if the mqtt topic contains a +, we replace it with @singleLevelWildcardRegex
+            // if the mqtt topic contains a #, we replace it with @multiLevelWildcardRegex
+            // if the mqtt topic contains a placeholder (pattern \{\w+\}), we replace it with @placeholderRegex
+            String regex = rule.getMqttTopicPattern().replaceAll("\\{\\w+\\}", PLACEHOLDER_REGEX)
```
kyguy, line 51: Could we abstract the hardcoded regex strings as well?

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,68 @@
...
+                    .replace("#", MULTIPLE_LEVEL_WILDCARD_REGEX).replace("+", SINGLE_LEVEL_WILDCARD_REGEX);
+            if (mqttTopic.matches(regex)) {
+                String mappedKafkaTopic = rule.getKafkaTopicTemplate();
+                String[] mqttTopicPatternParts = rule.getMqttTopicPattern().split("/");
+                for (String placeholderKey : mqttTopicPatternParts) {
+                    if (placeholderKey.matches("\\{\\w+\\}")) {
```
kyguy, line 57: Could we abstract the hardcoded regex strings as well?

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
```diff
@@ -0,0 +1,103 @@
...
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        mappingRules.sort(Comparator.comparing(MappingRule::getMqttTopicPatternLevels).reversed());
```
kyguy, line 50: Clever :)

ppatierno: I don't think this is enough to have the MQTT pattern with `+` taking precedence on the ones with `#` at the end.
If I got it right, this comparator is based on counting how many `/` are in the pattern (so what you call "levels") but ... take these two MQTT patterns as an example:

`/foo/+/bar/+` and `/foo/+/bar/#`

They have the same "levels" (same number of  `/` character) but ...

If the producer is sending to `/foo/10/bar/5` we want it being matched with `/foo/+/bar/+` and not `/foo/+/bar/#`. How we are sure that with the comparator you put in place we have this order?

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
ppatierno: Totally agree with Kyle. On class creation we should "map" the rules from JSON to our internal list of regex.
It would be better to use the `Pattern` class and the `Pattern.compile` static method. 
You could "compile" each regex and having the corresponding list of `Pattern` instances.
This way the matching operation would be even faster.

So you could have a `List<Pattern> patterns` field and filling it going through the rules, compiling each regex.

```java
for (MappingRule rule : this.rules) {
    String regex = rule.getMqttTopicPattern().replaceAll("\\{\\w+\\}", PLACEHOLDER_REGEX)
                      .replace("#", MULTIPLE_LEVEL_WILDCARD_REGEX).replace("+", SINGLE_LEVEL_WILDCARD_REGEX);
    patterns.add(Pattern.compile(regex));
}
```

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Responsible for handling all the topic mapping.
+ *
+ * @see MappingRule
+ */
+public class MqttKafkaMapper {
+
+    private static String DEFAULT_KAFKA_TOPIC = "messages_default";
+    private ArrayList<MappingRule> rules;
```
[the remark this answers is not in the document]
ppatierno: +1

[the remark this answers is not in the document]
ppatierno: Also please use interfaces for collections field declaration, so `List<MappingRule>`.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,68 @@
...
+    // matches any character except line terminators. Used to replace the placeholders with {something} in the mqtt pattern.
+    private static String PLACEHOLDER_REGEX = "(.*)";
+    // matches any character except a forward slash (/). Used to replace the + in the mqtt pattern.
+    private static String SINGLE_LEVEL_WILDCARD_REGEX = "[^/]+";
+    // matches any character after the string. Used to replace the # in the mqtt pattern.
+    private static String MULTIPLE_LEVEL_WILDCARD_REGEX = ".*";
```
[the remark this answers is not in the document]
ppatierno: +1 ... all constants have to be `final`

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
ppatierno: Use `List<MappingRule>`

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -25,6 +25,7 @@
  */
 public class MqttServerHandler extends SimpleChannelInboundHandler<MqttMessage> {
     private static final Logger logger = LoggerFactory.getLogger(MqttServerHandler.class);
+    private MqttKafkaMapper mapper = new MqttKafkaMapper(new ArrayList<>());
```
ppatierno, line 28: You are initializing the mapper with an empty list of rules, and I don't see them to be loaded later. How is this going to work?

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
ppatierno: +1

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRule.java
ppatierno: +1

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
+ * Unit tests for {@link MqttKafkaMapper}
+ */
+public class MqttKafkaMapperTest {
+
+    private ArrayList<MappingRule> mappingRules;
```
ppatierno, line 25: Still using `List` interface

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
ppatierno: But I also see you have the JSON file in the `resources` folder, why not using it instead of this big String?

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
ppatierno: Maybe something like this is more meaninful ...

```java
mappingRules = mapper.readerForListOf(MappingRule.class).readValue(TOPIC_MAPPING_RULES_JSON);
```

ppatierno (changes requested):
I had a first pass ... more to come while testing the code ;-)

ppatierno:
Other than my specific comments, I think that we should try to narrow down tests in simpler ones with less rules for each test and not using a bigger JSON file loading them all.
I mean, I would like to have tests like this:

```java
@Test
    public void test() {
        List<MappingRule> rules = new ArrayList<>();
        rules.add(new MappingRule("fleet/{fleet}/vehicle/{vehicle}", "fleet_{fleet}"));
        MqttKafkaMapper mapper = new MqttKafkaMapper(rules);
        assertThat(mapper.map("fleet/5/vehicle/10"), is("fleet_5"));
    }
```

You create your rules, provide them to a mapper and then use/test it.
In this way, we can have more tests focusing on specific patterns we want to test. You see the rules defined in the test, without the need to move to the JSON file/rules. You can immediately see what expecting.

antonio-pedro99:
@ppatierno I totally agree with you.

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
ppatierno: Well loading from a file in the resources folder is not a complex thing :-)
It's not about configuring it's just about doing a `getResourceAsStream` and loading it.
Anyway I added a more general comment about not using the JSON resources file at all in our tests, we should be more specific and explicit.

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRule.java
ppatierno: Yeah I was thinking the same while reading the code again. I agree with Antonio.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
ppatierno: Let's see how it is :-)

antonio-pedro99:
> Other than my specific comments, I think that we should try to narrow down tests in simpler ones with less rules for each test and not using a bigger JSON file loading them all. I mean, I would like to have tests like this:
> 
> ```java
> @Test
>     public void test() {
>         List<MappingRule> rules = new ArrayList<>();
>         rules.add(new MappingRule("fleet/{fleet}/vehicle/{vehicle}", "fleet_{fleet}"));
>         MqttKafkaMapper mapper = new MqttKafkaMapper(rules);
>         assertThat(mapper.map("fleet/5/vehicle/10"), is("fleet_5"));
>     }
> ```
> 
> You create your rules, provide them to a mapper and then use/test it. In this way, we can have more tests focusing on specific patterns we want to test. You see the rules defined in the test, without the need to move to the JSON file/rules. You can immediately see what expecting.

in addtion, if we want to test multiple entries. we just more MappingRule to the rules list.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
antonio-pedro99: @ppatierno is there any reason we should use List instead of array list?

ppatierno:
> in addtion, if we want to test multiple entries. we just add more MappingRule to the rules list.

of course, mine was just an example. For sure we want tests where there are multiple rules.

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
antonio-pedro99: @kyguy sorry, I did not get the question.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
ppatierno: > the mapper is not meant to be used withing the server yet.

So I would remove this changes to the server for now. Let's make only changes which are meaningful to the PR.

> is there any reason we should use List instead of array list?

In inheritance, you usually want always have the reference to be declared as an interface or base/abstract class. If you want to move to an ArrayList to LinkedList for example it's much simpler. Use a concrete class as declaration only if you don't have an interface or base class. In any case you can cast if needed.

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
antonio-pedro99: If you are referring to line 46, I think the IDE's code reformat did it.

antonio-pedro99: @ppatierno magic.

ppatierno: I think Kyle was asking why the `});` is not at the end of line 45.
I think this is just the IDE because you are declaring an anonymous class and IDE just moves the closed parenthesis to the next line. Anyway I would follow my suggestion instead of having this anonymous class.

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,68 @@
...
+            String regex = rule.getMqttTopicPattern().replaceAll("\\{\\w+\\}", PLACEHOLDER_REGEX)
+                    .replace("#", MULTIPLE_LEVEL_WILDCARD_REGEX).replace("+", SINGLE_LEVEL_WILDCARD_REGEX);
+            if (mqttTopic.matches(regex)) {
+                String mappedKafkaTopic = rule.getKafkaTopicTemplate();
+                String[] mqttTopicPatternParts = rule.getMqttTopicPattern().split("/");
+                for (String placeholderKey : mqttTopicPatternParts) {
+                    if (placeholderKey.matches("\\{\\w+\\}")) {
+                        int index = Arrays.asList(mqttTopicPatternParts).indexOf(placeholderKey);
+                        String placeholder = mqttTopic.split("/")[index];
+                        mappedKafkaTopic = mappedKafkaTopic.replace(placeholderKey, placeholder);
```
ppatierno, line 60: So this replacement is not going to work well, you need a more sophisticated logic, at least checking the placeholder names. 
Check this test:

```java
@Test
    public void test() {
        List<MappingRule> rules = new ArrayList<>();
        rules.add(new MappingRule("fleet/{fleet}/vehicle/{vehicle}", "fleet_{flee}"));
        MqttKafkaMapper mapper = new MqttKafkaMapper(rules);
        assertThat(mapper.map("fleet/5/vehicle/10"), is("messages_default"));
    }
```

Why I am expecting to be "message_default"? Because I made a mistake on purpose on the Kafka template ... notice it's `fleet_{flee}` and not `fleet_{fleet}` ... the placeholder name is wrong.
You are just replacing placeholders with the values got from the MQTT pattern but if a placeholder doesn't exist as in this case, it won't be replaced.
In this test, the output Kafka topic is just ... the Kafka template, so it's wrong.

```shell
Expected: is "messages_default"
     but: was "fleet_{flee}"
```

ppatierno (changes requested):

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
antonio-pedro99: @ppatierno I think so, I have fixed that.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: @kyguy yes, sounds good.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: @ppatierno I am now using List everywhere I was using arraylist.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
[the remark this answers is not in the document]
antonio-pedro99: @kyguy done

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: @ppatierno you can take a look now.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: Yes, I was thinking about it while implementing. 

In the beginning, I was taking all the placeholders and put them in a hash Map (placeholder name as the key, and the corresponding value from the MQTT topic (input) as the value) and then build the Kafka topic using the Kafka topic template and the hash Map. But the behavior is the same if the key is not found. 

As this should be user defined, I think we can just check the names, if it does not match, we can just return an exception: mapping failed because x was y. 

I can still use the HashMap idea.

antonio-pedro99: or maybe, I can check if all the placeholders in the template and in the mqtt pattern are same, if not we can return an error before starting the mapping.

antonio-pedro99: I did not use a HashMap.
Just checked if the placeholder in the MQTT pattern is also in the Kafka template. 
I have added tests as well.

antonio-pedro99: Just went through the code again and check some issues. I am working on it.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,134 @@
...
+     */
+    private String getKafkaTopicDelimiter(String kafkaTopicTemplate) {
+        if (kafkaTopicTemplate.contains("_")) {
+            return "_";
+        } else if (kafkaTopicTemplate.contains("-")) {
+            return "-";
+        } else if (kafkaTopicTemplate.contains(".")) {
+            return ".";
+        } else {
+            throw new IllegalArgumentException("The kafka topic template must contain a valid delimiter.");
```
ppatierno, line 131: I don't see why the topic template cannot be something like `foo{myvar}` without any delimiter.

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
```diff
@@ -0,0 +1,110 @@
...
+    public void testIllegalPlaceholder(){
+
+        List<MappingRule> rules = new ArrayList<>();
+        rules.add(new MappingRule("fleet_{fleet}", "fleet/{flee}/vehicle/{vehicle}"));
+
+        MqttKafkaMapper mapper = new MqttKafkaMapper(rules);
+
+        Exception exception = assertThrows(IllegalArgumentException.class, () -> {
+            mapper.map("fleet/4/vehicle/23");
+        });
```
ppatierno, line 77: I am not sure this should return an exception. In my mind when a match is not found we should use the default topic. Or maybe better, it has to be something configurable by the user.

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
```diff
@@ -0,0 +1,110 @@
...
+     * E.g. if the Topic Mapping Rules is empty, the default topic is used by default.
+     */
+    @Test
+    public void testDefaultTopic() {
+        List<MappingRule> rules = new ArrayList<>();
+
+        MqttKafkaMapper mapper = new MqttKafkaMapper(rules);
+
+        assertThat("Should use the default topic when no mapping pattern matches.",
+                mapper.map("sensor/temperature"), is(MqttKafkaMapper.DEFAULT_KAFKA_TOPIC));
```
ppatierno, line 37: well that's an example of just no rules ... but what when there are some rules and they don't match, for any reasons? see my next comment related to the wrong placeholder name, that could be another use case.

antonio-pedro99: The behavior would be just the same. No matches will be found.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: instead of delimiters will use regex.

antonio-pedro99: There would be one more update.
For now, I am just assuming the placeholders are after / in the MQTT topic, but there can be cases the topics will look like, `user{id}/info/{myvar}`.

ppatierno: I would say that this would not be a good MQTT pattern, so users could be making a mistake. Let's focus on having placeholders right after the `/` for now.

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
ppatierno: Of course `fleet{number}` is not a good Kafka topic if the MQTT pattern is `fleet/#` but if the pattern is `fleet/{number}` and the user sends to `fleet/10`, then `fleet10` will be a valid Kafka topic.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: In my r

> I don't see why the topic template cannot be something like `foo{myvar}` without any delimiter.

In my recent commit, I am handling this case, and it is working. Please take a look.
Tests like 
 `rules.add(new MappingRule("term{number}", "term/{number}"));`
 
`assertThat("Mqtt pattern term/{number} should be mapped to term{number}",
                mapper.map("term/4"), is("term4"));`
 were added.

ppatierno:
So, I took a look at the logic but I think it could be improved and maybe "simplified". I don't have the implementation but the overall picture of how it could work, so I will leave you @antonio-pedro99 to check how it could work and developed.
Let me use some examples to explain the logic I am envisaging.

From an MQTT perspective we could have the following three patterns:

```
buildings/{building}/rooms/{room}/device/{device}
buildings/+/rooms/+/device/+
buildings/+/rooms/{room}/device/+
```

So using all placeholders, all `+` or a mix of them.

When the user has placeholders in the MQTT pattern, they can be used in the Kafka topic template, so that we have a relationship between placeholder names.
When using the `+` in the MQTT pattern, I would avoid the user to use placeholders in the Kafka topic template because the "positional" replacement could be problematic.
What I am saying is ...

`buildings/{building}/rooms/{room}/device/{device}` --> `buildings_{building}_rooms_{room}/device_{device}` IS OK
`buildings/+/rooms/+/device/+` --> `buildings_{building}_rooms_{room}/device_{device}` IS NOT OK
`buildings/+/rooms/+/device/+` --> `buildings_data` IS OK
`buildings/+/rooms/{room}/device/+` --> `buildings_data_{room} IS OK

Also the Java Pattern(s) to the corresponding MQTT ones could be like following

`buildings/{building}/rooms/{room}/device/{device}` --> `buildings/(?<building>[^/]+)/rooms/(?<room>[^/]+)/device/(?<device>[^/]+)`
`buildings/+/rooms/+/device/+` --> `buildings/([^/]+)/rooms/([^/]+)/device/([^/]+)`
`buildings/+/rooms/{room}/device/+` --> `buildings/([^/]+)/rooms/(?<room>[^/]+)/device/([^/]+)`

As you can see I am defining regex groups with names when there is a placeholder so that I can access the group by name (see `matcher.group("nameofgroup")` method) when I want to get the placeholder value from the actual MQTT topic and replacing inside the Kafka topic template.

Said that, I think that the following are things you could think about:

- given the Kafka topic template with placeholders and MQTT topic pattern with placeholders, you have to be check that Kafka placeholders are in the list of the MQTT placeholders (in order to be able to do the replace with actual values).
- to build the Java Pattern(s) with named groups you need anyway to extract the list of placeholders from the MQTT topic pattern. Then it should be easy to format a group as `(?<name of the placeholder>[^/]+)` and replace to build the Java Pattern.
- when an actual MQTT topic matches a pattern, you can walk through the named groups, getting the value and replace into the Kafka topic template, to make the actual Kafka topic.

antonio-pedro99:
@ppatierno in this case
`buildings/+/rooms/+/device/+ --> buildings_{building}_rooms_{room}/device_{device}` IS NOT OK
what should be the output? Does not use exception suffices the problem?
In my recent approach, I am tackling this issue. 

Splitting the template and the MQTT pattern won't work because the template can have delimiters or not. So what I am doing is, at first getting all the placeholders from the template and add in hash Map with null value, in this case:
building:null,
room: null,
device: null.

Now after splitting the MQTT pattern to check for placeholders, if no placeholders that are in the hash Map are found, we can not build the actual Kafka topic from the template. Otherwise, we just update the value of the placeholders that are in the hash Map with their actual values from the MQTT topic (not the pattern). To build the actual Kafka topic, we use all the values from the hash map, ignoring all entries with null values.

antonio-pedro99:
@ppatierno yes, it is good idea to use named groups.

antonio-pedro99:
`1. given the Kafka topic template with placeholders and MQTT topic pattern with placeholders, you have to be check that Kafka placeholders are in the list of the MQTT placeholders (in order to be able to do the replace with actual values).`

please check my comment

ppatierno:
> what should be the output? Does not use exception suffices the problem?

I am not sure about exception, maybe we could log a warning and getting the default topic.

For the rest, it's simpler to see the code than reading the explanation :-)

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRule.java
```diff
@@ -0,0 +1,81 @@
...
+    public MappingRule() {
+    }
+
+    /**
+     * Constructor for MappingRule.
+     *
+     * @param kafkaTopicTemplate the kafka topic template.
+     * @param mqttTopicPattern   the mqtt topic pattern.
+     */
+    public MappingRule(String kafkaTopicTemplate, String mqttTopicPattern) {
```
ppatierno, line 38: My mental model is MQTT -> Kafka so isn't it better swapping the two parameters and having them in the opposite order?

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,161 @@
...
+                }
+                if (!part.equals(mqttTopicPatternParts[lastElementIndex])) {
+                    ruleRegex.append("/");
+                }
+            }
+            patterns.add(Pattern.compile(ruleRegex.toString()));
+        }
+
+        // add the regex for the placeholders in the end of the patterns list.
+        patterns.add(Pattern.compile(MQTT_TOPIC_PLACEHOLDER_REGEX));
```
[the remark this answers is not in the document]
antonio-pedro99: private final Pattern mqttTopicPlaceholderPattern = Pattern.compile(MQTT_TOPIC_PLACEHOLDER_REGEX); would work?

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,161 @@
...
+            if (matcher.matches()) {
+                String mappedKafkaTopic = rule.getKafkaTopicTemplate();
+                String[] mqttTopicPatternParts = rule.getMqttTopicPattern().split("/");
+
+                // find MQTT_TOPIC_PLACEHOLDER_REGEX in the kafkaTopicTemplate.
+                Matcher placeholderMatcher = patterns.get(patterns.size() - 1).matcher(rule.getKafkaTopicTemplate());
+                while (placeholderMatcher.find()) {
+                    placeholders.put(placeholderMatcher.group(), null);
+                }
+
```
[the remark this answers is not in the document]
antonio-pedro99: well pointed! Thanks.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,161 @@
...
+                } else if (part.equals(MQTT_TOPIC_SINGLE_LEVEL_WILDCARD_CHARACTER)) {
+                    ruleRegex.append(SINGLE_LEVEL_WILDCARD_REGEX);
+                } else if (part.equals(MQTT_TOPIC_MULTI_LEVEL_WILDCARD_CHARACTER)) {
+                    ruleRegex.append(MULTIPLE_LEVEL_WILDCARD_REGEX);
+                } else {
+                    ruleRegex.append(part);
+                }
+                if (!part.equals(mqttTopicPatternParts[lastElementIndex])) {
+                    ruleRegex.append("/");
+                }
```
ppatierno, line 129: I found a bug ... try a test with this rule:

```java
rules.add(new MappingRule("buildings_{building}_rooms_{room}_device_{device}", "buildings/+/rooms/+/device/+"));

assertThrows(IllegalArgumentException.class, () -> mapper.map("buildings/10/rooms/5/device/3"));
```

it doesn't add `/` after finding a `+` because ... `+` is the last element :-)

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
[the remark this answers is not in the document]
kyguy: This still needs love

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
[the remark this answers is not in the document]
kyguy: This still needs love too

Maybe the commit that carried the corrections hasn't been pushed yet?

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,68 @@
...
+    // matches any character except line terminators. Used to replace the placeholders with {something} in the mqtt pattern.
+    private static String PLACEHOLDER_REGEX = "(.*)";
+    // matches any character except a forward slash (/). Used to replace the + in the mqtt pattern.
+    private static String SINGLE_LEVEL_WILDCARD_REGEX = "[^/]+";
+    // matches any character after the string. Used to replace the # in the mqtt pattern.
+    private static String MULTIPLE_LEVEL_WILDCARD_REGEX = ".*";
+
+    /**
+     * Constructor
+     * <p>
```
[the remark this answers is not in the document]
kyguy: Same here

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,161 @@
...
+        for (MappingRule rule : this.rules) {
+            Matcher matcher = this.patterns.get(this.rules.indexOf(rule)).matcher(mqttTopic);
+            HashMap<String, String> placeholders = new HashMap<>();
+
+            if (matcher.matches()) {
+                String mappedKafkaTopic = rule.getKafkaTopicTemplate();
+                String[] mqttTopicPatternParts = rule.getMqttTopicPattern().split("/");
+
+                // find MQTT_TOPIC_PLACEHOLDER_REGEX in the kafkaTopicTemplate.
+                Matcher placeholderMatcher = patterns.get(patterns.size() - 1).matcher(rule.getKafkaTopicTemplate());
```
kyguy, line 75: Would it be simpler if  we defined another field like this: 
```
private final static Pattern MQTT_TOPIC_PLACEHOLDER_PATTERN = Pattern.compile(MQTT_TOPIC_PLACEHOLDER_REGEX);
```
and then did this:
```suggestion
                Matcher placeholderMatcher = MQTT_TOPIC_PLACEHOLDER_PATTERN.matcher(rule.getKafkaTopicTemplate());
```
Maybe I am just complicating things, let me know what you think!

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,161 @@
...
+                    }
+                }
+
+                for (Map.Entry<String, String> entry : placeholders.entrySet()) {
+                    if (entry.getValue() != null) {
+                        mappedKafkaTopic = mappedKafkaTopic.replace(entry.getKey(), entry.getValue());
+                    } else {
+                        throw new IllegalArgumentException("The placeholder " + entry.getKey() + " was not assigned any value.");
+                    }
+                }
```
kyguy, line 92: Would it make sense to scan for missing placeholder values during the initial scan of the MappingRules when we are building our Pattern list and throw the exception then?

kyguy (commented):
Great progress Antonio!

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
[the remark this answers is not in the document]
antonio-pedro99: @kyguy this has been fixed.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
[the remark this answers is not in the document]
antonio-pedro99: This has already been fixed :)

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: @kyguy I have already made this change.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: @kyguy let me think about.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,161 @@
...
+            }
+
+        }
+        return DEFAULT_KAFKA_TOPIC;
+    }
+
+    /**
+     * Helper method for Building the regex expressions for the mapping rules.
+     */
+    private void buildRegex() {
```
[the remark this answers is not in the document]
antonio-pedro99: buildRegex is only used inside the Mapper class, I do not think it would make much sense.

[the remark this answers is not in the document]
antonio-pedro99: @kyguy what I was thinking is creating an helper class with all these methods not directly related to the mapper. Something like

`class MqttKafkaMapperHelper {
     //all regex constants here.
     buildRegex() and other methods not directly related to the mapper
}`
Doing so, the MqttKafkaMapper could have only the map() method.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: interesting...

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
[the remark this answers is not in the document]
antonio-pedro99: fixed.

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRule.java
antonio-pedro99: fixed

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: fixed.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,161 @@
...
+     * Maps an MQTT topic to a Kafka topic. The topic is mapped according to the defined mapping rules.
+     *
+     * @param mqttTopic represents an MQTT topic.
+     * @return a valid Kafka topic.
+     * @see MappingRule
+     */
+    public String map(String mqttTopic) {
+        for (MappingRule rule : this.rules) {
+            Matcher matcher = this.patterns.get(this.rules.indexOf(rule)).matcher(mqttTopic);
+            HashMap<String, String> placeholders = new HashMap<>();
```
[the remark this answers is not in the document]
antonio-pedro99: fixed

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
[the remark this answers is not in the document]
ppatierno: I would stay with the current implementation we have now.
Moving from rules to patterns is an internal stuff to the mapper class which cannot deal with rules but needs pattern for matching topics.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,161 @@
...
+     * @see MappingRule
+     */
+    public String map(String mqttTopic) {
+        for (MappingRule rule : this.rules) {
+            Matcher matcher = this.patterns.get(this.rules.indexOf(rule)).matcher(mqttTopic);
+
+            if (matcher.matches()) {
+                HashMap<String, String> placeholders = new HashMap<>();
+                String mappedKafkaTopic = rule.getKafkaTopicTemplate();
+                String[] mqttTopicPatternParts = rule.getMqttTopicPattern().split("/");
```
ppatierno, line 73: this could be improved ... the idea is, using the placeholder pattern to match only the placeholders in the MQTT pattern instead of bringing with you pieces which are not placeholder but you still check in the next two loops.
For example, having `/buildings/{building}/rooms/{room}`, if you just split you will get the array will all 4 parts `[buildings, {building}, rooms, {room}]` while what you actually needs is just `[{building}, {room}]` with just placeholders. This way you don't have to loop across useless element in the array in the next two loops.

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Test the mapping of single level topics.
+     */
+    @Test
+    public void testIllegalPlaceholder() {
+
+        List<MappingRule> rules = new ArrayList<>();
+        rules.add(new MappingRule("fleet/{flee}/vehicle/{vehicle}", "fleet_{fleet}"));
+        rules.add(new MappingRule("buildings/+/rooms/+/device/+", "buildings_{building}_rooms_{room}_device_{device}"));
```
ppatierno, line 78:
```suggestion
        rules.add(new MappingRule("buildings/+/rooms/+/device/+", "buildings_{building}_rooms_{room}_device_{device}"));
        rules.add(new MappingRule("building/{building}/room/{room}", "building_{building}_room_{room}_{noexistingplaceholder}"));
```

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapperTest.java
```diff
@@ -0,0 +1,121 @@
...
+                mapper.map("fleet/4/vehicle/23/velocity"), is("fleet_23"));
+
+        assertThat("Mqtt pattern building/{building}/room/{room}/# should be mapped to building_{building}_room_{room}",
+                mapper.map("building/4/room/23/temperature"), is("building_4_room_23"));
+
+        assertThat("Mqtt pattern building/{building}/# should be mapped to building_{building}_others",
+                mapper.map("building/405/room"), is("building_405_others"));
+
+        assertThat("Mqtt pattern building/# should be mapped to building_others",
+                mapper.map("building/101"), is("building_others"));
```
ppatierno, line 119:
```suggestion
                mapper.map("building/101"), is("building_others"));
                assertThrows(IllegalArgumentException.class, () ->
                mapper.map("building/10/room/5"));
```

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -0,0 +1,161 @@
...
+                        if (placeholderKey.matches(MQTT_TOPIC_PLACEHOLDER_REGEX) && placeholders.containsKey(placeholderKey)) {
+                            placeholders.put(placeholderKey, matcher.group(removeBrackets(placeholderKey)));
+                        }
+                    }
+
+                    for (Map.Entry<String, String> entry : placeholders.entrySet()) {
+                        if (entry.getValue() != null) {
+                            mappedKafkaTopic = mappedKafkaTopic.replace(entry.getKey(), entry.getValue());
+                        } else {
+                            throw new IllegalArgumentException("One or more placeholders were not assigned any value.");
```
ppatierno, line 92: We can be precise about which entry doesn't have a value. Of course, there could be more but because we are throwing an exception and failing, we are not moving next (I mean we are not collecting ALL the entries without a value). So I would say let's be specific in the exception message to show which placeholder didn't get a value so the user knows exactly what to fix.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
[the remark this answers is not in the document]
antonio-pedro99: Would it not be cleaner if the mapper would just deal with Mapping?

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: makes sense!

antonio-pedro99: I was just doing it.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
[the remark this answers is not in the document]
ppatierno: the mapper cannot work without the logic of building patterns from rules, it's part of it. Having a different class just for that and making a dependency doesn't make any sense to me imho.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
```diff
@@ -69,20 +65,25 @@ public String map(String mqttTopic) {
 
             if (matcher.matches()) {
                 HashMap<String, String> placeholders = new HashMap<>();
+                Pattern placeholderPattern = Pattern.compile(MQTT_TOPIC_PLACEHOLDER_REGEX);
```
ppatierno, line 68: while compiling the pattern every time you run the `map` method and there is a match?
Just compile it one time in the constructor and you are happy :-)
In the end it's like to be part of the `buildRegex` you already call there.

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttKafkaMapper.java
antonio-pedro99: Marked as fixed.

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRule.java
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Constructor for MappingRule.
+     *
+     * @param kafkaTopicTemplate the kafka topic template.
+     * @param mqttTopicPattern   the mqtt topic pattern.
+     */
+    public MappingRule(String mqttTopicPattern, String kafkaTopicTemplate) {
+        this.kafkaTopicTemplate = kafkaTopicTemplate;
+        this.mqttTopicPattern = mqttTopicPattern;
```
kyguy, line 39: Personally would find it more intuitive if we swapped the position of `kafkaTopicTemplate` and `mqttTopicPattern` so that `mqttTopicPattern` comes first (Going from MQTT topic -> Kafka topic)

Same for the comment for this method.

Same for the comment for the class.

kyguy (approved):
Solid work @antonio-pedro99, left a small nit, feel free to ignore

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRule.java
kyguy, line 39: Like this:
```suggestion
        this.mqttTopicPattern = mqttTopicPattern;
        this.kafkaTopicTemplate = kafkaTopicTemplate;
```

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRule.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * Constructor for MappingRule.
+     *
+     * @param kafkaTopicTemplate the kafka topic template.
+     * @param mqttTopicPattern   the mqtt topic pattern.
+     */
+    public MappingRule(String mqttTopicPattern, String kafkaTopicTemplate) {
+        this.kafkaTopicTemplate = kafkaTopicTemplate;
```
kyguy, line 38: Like this
```suggestion
        this.mqttTopicPattern = mqttTopicPattern;
        this.kafkaTopicTemplate = kafkaTopicTemplate;
```

antonio-pedro99: fixed.

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRule.java
antonio-pedro99: fixed

ppatierno (approved):
LGTM. Thanks for the hard work and effort you put into this ;-)

PR #8, opened 2023-06-23
Setup config server and load topic mapping rules from a file.

antonio-pedro99:
This PR add the following changes:

>  Ability to configure the basic information about the Bridge
>  Ability to load the application.properties
>  Added two command lines arguments:
--config-file and --mapping-rules
`--config-file` is the path name of the application properties file.
and `--mapping-rules` is the path name of the topic mapping rules file. This should be a .json file.

For all the changes, tests were added.


src/main/java/io/strimzi/kafka/bridge/mqtt/config/MqttConfig.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public static final String MQTT_HOST = MQTT_CONFIG_PREFIX + "host";
+
+    public static final String MQTT_PORT = MQTT_CONFIG_PREFIX + "port";
+
+    public static final String DEFAULT_MQTT_HOST = "localhost";
+
+    public static final int DEFAULT_MQTT_PORT = 1883;
+
+    public static String MAPPER_RULES_PATH;
```
kyguy, line 27: Is the plan to store the mapper rules in this variable? I don't think this variable is being used right now

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -24,6 +29,21 @@
...
     private static final Logger logger = LoggerFactory.getLogger(MqttServerHandler.class);
+    private MqttKafkaMapper mqttKafkaMapper;
+
+    /**
+     * Constructor
+     */
+    public MqttServerHandler() {
+        super(false);
+        try {
+            MappingRulesLoader mappingRulesLoader = MappingRulesLoader.getInstance();
```
kyguy, line 40: Not sure I understand the purpose `.getInstance()` method, couldn't we just create a MappingRulesLoader instance by doing:
```suggestion
            MappingRulesLoader mappingRulesLoader = new MappingRulesLoader();
```

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
```diff
@@ -0,0 +1,64 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * Helper class to Load the rules from the configuration file
+ */
+public class MappingRulesLoader {
+
+    private static final MappingRulesLoader instance = new MappingRulesLoader();
```
kyguy, line 19: Not sure I understand we we create a new instance here in this class, why not just rely on the constructor?

src/test/resources/mapping-rules.json
```diff
@@ -0,0 +1,30 @@
+[
+  {
+    "kafkaTopic": "building_{building}_room_{room}",
+    "mqttTopic": "building/{building}/room/{room}/#"
```
kyguy, line 4: Personally would swap the positions of kafkaTopic and mqttTopic like we did in the last PR

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * Get the singleton instance of the MappingRulesLoader
+     *
+     * @return the singleton instance of the MappingRulesLoader
+     */
+    public static MappingRulesLoader getInstance() {
+        return instance;
+    }
```
kyguy, line 36: Are you are following singleton pattern to ensure only one instance of MappingRulesLoader is created?

Is it really necessary for this case?

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
```diff
@@ -0,0 +1,64 @@
...
+        BufferedReader reader = new BufferedReader(new FileReader(this.mapperRuleFilePath));
+        String line;
+        // build the JSON string
+        while ((line = reader.readLine()) != null) {
+            json.append(line);
+        }
+        // deserialize the JSON array to a list of MappingRule objects
+        return mapper.readValue(json.toString(), mapper.getTypeFactory().constructCollectionType(List.class, MappingRule.class));
+    }
+}
```
kyguy, line 64: Any strong reason to remove this class and just  just move this method to the MqttConfig class?

kyguy (commented):
Awesome work @antonio-pedro99, just left some minor comments and had a couple format questions but overall  looks great!

src/main/java/io/strimzi/kafka/bridge/mqtt/Main.java
```diff
@@ -5,23 +5,88 @@
 package io.strimzi.kafka.bridge.mqtt;
 
 import io.netty.channel.ChannelOption;
+import io.netty.channel.EventLoopGroup;
 import io.netty.channel.nio.NioEventLoopGroup;
+import io.strimzi.kafka.bridge.mqtt.config.BridgeConfig;
+import io.strimzi.kafka.bridge.mqtt.config.ConfigRetriever;
+import io.strimzi.kafka.bridge.mqtt.utils.MappingRulesLoader;
 import io.strimzi.kafka.bridge.mqtt.core.MqttServer;
+import org.apache.commons.cli.*;
```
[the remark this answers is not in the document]
antonio-pedro99: Sorry, I forgot to check that,

src/main/java/io/strimzi/kafka/bridge/mqtt/config/MqttConfig.java
antonio-pedro99: yes, we are not using, I forgot to remove it.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
antonio-pedro99: No, the MappingRulesLoader is responsible for loading the rules from a file. It needs the file path. This path is set only once during the bridge startup(--mapping-rules=path/to/file).

Please take a look at [this line](https://github.com/strimzi/strimzi-mqtt-bridge/pull/8/files#:~:text=MappingRulesLoader.getInstance().setMapperRuleFilePath(mappingRulesFile)%3B)
I really had a harsh time finding the best way to adopt this.

But here are some reasons to adopt singleton in this scenario.
1. I do not want to pass the rules path from the Main class to other classes. This path is provided in the args --mapping-rules=path/to/file, if I give the responsibility to load the mapping rules file to MqttConfig, I will have to pass the loaded rules as an argument to the MqttServer class, then to the MqttServerInitializer class and finally to the MqttServerHandler class which has the MqttKafkaMapper instance. So it would be hard to maintain and track the lifecycle of the mapping rules.

2. Single responsibility principle, I want to make the project clear and easy maintainable. For the entire lifecycle of the bridge, we just need one instance of the MappingRulesLoader, set the path from the very beginning of the bridge startup and use the loadRules() anywhere in the project.

3. Again single responsibility, right now if news rules are added while the bridge is already running, we won't get the fresh rules, to get it, we need to restart the server. Even if we are not considering this case now, I thought it would be better to implement something that will make it easy to change the scenario. For example, in future we may want to keep track of the mapping rules file, check for new entries and update the rules inside the bridge, we would just add all the logic in the MappingRulesLoader.

So to answer the question, if we do
MappingRulesLoader = new MappingRulesLoader();
we will need to set the path which would come from the Main class.

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
antonio-pedro99: I have answered [here](https://github.com/strimzi/strimzi-mqtt-bridge/pull/8#discussion_r1240604882)

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
antonio-pedro99: I have answered [here](https://github.com/strimzi/strimzi-mqtt-bridge/pull/8#discussion_r1240604882)

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
antonio-pedro99: Still, I have answered [here](https://github.com/strimzi/strimzi-mqtt-bridge/pull/8#discussion_r1240604882)

config/mapping-rules.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  {
+    "mqttTopic": "sensors/#",
+    "kafkaTopic": "sensor_others"
+  },
+  {
+    "mqttTopic": "building/#",
+    "kafkaTopic": "building_others"
+  }
+]
```
ppatierno, line 30: we don't need this here, because it's a user specific file. While the application.properties file can have some defaults which we are setting, the rules is business/use case specific. We can provide an example somewhere but not in this folder where we expect something more "official".

pom.xml
```diff
@@ -72,5 +73,10 @@
             <version>${junit.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>commons-cli</groupId>
+            <artifactId>commons-cli</artifactId>
+            <version>${commons-cli.version}</version>
+        </dependency>
```
ppatierno, line 80: Can you move this dependency before the section marked as `<!-- Testing -->` please'

src/main/java/io/strimzi/kafka/bridge/mqtt/config/KafkaConfig.java
```diff
@@ -0,0 +1,74 @@
...
+public class KafkaConfig extends AbstractConfig {
+
+    // Prefix for all the specific configuration parameters for Kafka in the properties file
+    public static final String KAFKA_CONFIG_PREFIX = "kafka.";
+
+    // Kafka bootstrap server configuration. This will be removed when we add the apache kafka client
+    public static final String BOOTSTRAP_SERVERS_CONFIG = KAFKA_CONFIG_PREFIX + "bootstrap.servers";
+
+    private final KafkaProducerConfig kafkaProducerConfig;
+    private final KafkaAdminConfig kafkaAdminConfig;
```
ppatierno, line 23: Kafka admin? I guess this is coming from the Strimzi HTTP bridge, which is fine but we should remove everything we don't need like this Kafka admin related config.

src/main/java/io/strimzi/kafka/bridge/mqtt/config/KafkaProducerConfig.java
```diff
@@ -0,0 +1,49 @@
...
+ * Represents configurations related to Kafka
+ * @see AbstractConfig
+ */
+public class KafkaProducerConfig extends AbstractConfig {
+
+    // Prefix for all the specific configuration parameters for Kafka producer in the properties file
+    public static final String KAFKA_PRODUCER_CONFIG_PREFIX = KafkaConfig.KAFKA_CONFIG_PREFIX + "producer.";
+
+    // Acknowledgement level for Kafka producer
+    public static final String ACKS_LEVEL = KAFKA_PRODUCER_CONFIG_PREFIX + "acks";
```
ppatierno, line 20: we don't want a specific constant for ALL the possible Kafka producer config parameters. If you open the Kafka doc they are a lot but in any case it doesn't make sense. Everything coming after `kafka.producer.` is put inside the map in the base class and then passed to the Kafka producer constructor. You can just follow the Strimzi HTTP bridge pattern here, without adding this.

config/mapping-rules.json
antonio-pedro99: @ppatierno what about the mapping rules for testing?

pom.xml
antonio-pedro99: Yes!

src/main/java/io/strimzi/kafka/bridge/mqtt/config/KafkaConfig.java
antonio-pedro99: Sure!

src/main/java/io/strimzi/kafka/bridge/mqtt/config/KafkaProducerConfig.java
antonio-pedro99: @ppatierno in HTTP bridge some Kafka related parameter constants are from `org.apache.kafka.clients` package and I thought maybe I should not add any dependency related to Kafka yet. But if you are okay, I can go ahead.

config/mapping-rules.json
ppatierno: mapping rules for testing are ok, they can stay there.

src/main/java/io/strimzi/kafka/bridge/mqtt/Main.java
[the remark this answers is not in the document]
ppatierno: Yeah we don't have such a checkstyle right now here, but will add in the future. So it's better to avoid importing with wildcard as Kyle said.

src/main/java/io/strimzi/kafka/bridge/mqtt/config/KafkaProducerConfig.java
ppatierno: Why do you need Kafka clients dependency right now? You are not going to create a Kafka producer for now.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
ppatierno: While I am not totally against the idea of having a singleton pattern, I would avoid having a setter to provide the path because it would mean you could set it whenever through the code, while I would leave the singleton "configuration" immutable. I would have an `init` method (where you can provide the file path) which would raise and exception if called one more time.

src/main/java/io/strimzi/kafka/bridge/mqtt/config/KafkaProducerConfig.java
ppatierno: If you take a look at how the HTTP bridge works, all bridge parameters for the Kafka producer are prefixed with `kafka.producer.` and the last part has the name of a Kafka producer parameter.
For example, if you set `kafka.producer.acks` in the bridge configuration, the `acks` part is the name of a parameter for the underneath Kafka producer directly (see the Kafka documentation for that). You don't need a constant for it.

config/application.properties
```diff
@@ -0,0 +1,12 @@
...
+
+#MQTT server common
+mqtt.server.host=localhost
+mqtt.server.port=1883
+
+#Apache Kafka common
+kafka.bootstrap.servers=localhost:9092
+
+#Apache Kafka producer
+kafka.producer.acks=1
```
ppatierno, line 12: Other than the comment I made around the acks constant you declared in the Kafka configuration, I think that we should remove this for now from everywhere.
We have to think how to deal with making a relationship between the MQTT QoS and the Kafka acks.
It could even be something more dynamic and not static configured through the file.
But we should think more about it, not now.

src/test/resources/application.properties
```diff
@@ -0,0 +1,12 @@
...
+
+#MQTT server common
+mqtt.server.host=localhost
+mqtt.server.port=1883
+
+#Apache Kafka common
+kafka.bootstrap.servers=localhost:9092
+
+#Apache Kafka producer
+kafka.producer.acks=1
```
ppatierno, line 12: Same comment as before, about removing this parameter from the configuration.

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
ppatierno:
```suggestion
    private static final MappingRulesLoader INSTANCE = new MappingRulesLoader();
```

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
```diff
@@ -0,0 +1,64 @@
...
+    private MappingRulesLoader() {
+    }
+
+    /**
+     * Get the singleton instance of the MappingRulesLoader
+     *
+     * @return the singleton instance of the MappingRulesLoader
+     */
+    public static MappingRulesLoader getInstance() {
+        return instance;
```
ppatierno, line 35:
```suggestion
        return INSTANCE;
```

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    public List<MappingRule> loadRules() throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
+        StringBuilder json = new StringBuilder();
+        BufferedReader reader = new BufferedReader(new FileReader(this.mapperRuleFilePath));
+        String line;
+        // build the JSON string
+        while ((line = reader.readLine()) != null) {
+            json.append(line);
+        }
```
ppatierno, line 60: can you give it a try to this method? https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/ObjectMapper.html#readValue(java.io.File,%20java.lang.Class)

Instead of reading the file first, line by line, building a String and then readValue on that String, the above method could be able to get a `File` instance directly and loading directly into a list of `MappinRule`(s).

ppatierno: I just tried, you can replace everything with just one line:

```java
return mapper.readValue(Path.of(this.mapperRuleFilePath).toFile(), mapper.getTypeFactory().constructCollectionType(List.class, MappingRule.class));
```

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
antonio-pedro99: This is wise! :). Thanks for the suggestion.

antonio-pedro99: Should we do something like this for more clarity ?
` // get the file from the path
        final File mappingRulesJsonFile = Path.of(this.mapperRuleFilePath).toFile();
        
        final CollectionType collectionType = mapper.getTypeFactory().constructCollectionType(List.class, MappingRule.class);
        
        // deserialize the JSON array to a list of MappingRule objects
        return mapper.readValue(mappingRulesJsonFile, collectionType);`

config/application.properties
antonio-pedro99: should I remove KafkaProducerConfig from this PR?
I think it should be good to keep it.

src/main/java/io/strimzi/kafka/bridge/mqtt/config/KafkaProducerConfig.java
antonio-pedro99: (here)[https://github.com/strimzi/strimzi-kafka-bridge/blob/eab72b43d3b63742a2ffa3da2891235ab310564c/src/test/java/io/strimzi/kafka/bridge/ConfigRetrieverTest.java#L44] the ACK is coming from `import org.apache.kafka.clients.producer.ProducerConfig`

antonio-pedro99: but I have removed the constant! and just kept one prefix

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
antonio-pedro99: I have done that!

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
ppatierno: I don't think so, it's just a bunch of more lines of code and local variables, which don't make it clearer imho.

config/application.properties
ppatierno: Keep the `KafkaProducerConfig` class but remove this configuration line here for now.

antonio-pedro99: Done

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
```diff
@@ -0,0 +1,75 @@
...
+        if (!initialized) {
+            throw new IllegalStateException("MappingRulesLoader is not initialized");
+        }
+
+        ObjectMapper mapper = new ObjectMapper();
+
+        // get the file from the path
+        final File mappingRulesJsonFile = Path.of(this.mapperRuleFilePath).toFile();
+
+        final CollectionType collectionType = mapper.getTypeFactory().constructCollectionType(List.class, MappingRule.class);
```
ppatierno, line 70: As I mentioned in a previous comment I would avoid these lines and additional variables, they don't make the code clearer imho.

ppatierno (changes requested):

src/test/resources/application.properties
ppatierno: @antonio-pedro99 the acks configuration is still here in the test, we should remove from here as well for now.

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
antonio-pedro99: Sorry, it was a mistake, I forgot to remove from there. 
I have fixed most of the issues raised here.

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
antonio-pedro99: I have fixed that in my recent commit

src/test/resources/application.properties
antonio-pedro99: Yes, I forgot to remove. 
I have updated it. Please take a look.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
antonio-pedro99: I am waiting for @kyguy to say something regarding this conversation.

src/main/java/io/strimzi/kafka/bridge/mqtt/Main.java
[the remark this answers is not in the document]
antonio-pedro99: IJ made that :(. But I have fixed. Please take a look.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
kyguy: That makes sense @antonio-pedro99 good forethought, thanks for explaining!

src/test/java/io/strimzi/kafka/bridge/mqtt/core/ConfigRetrieverTest.java
```diff
@@ -0,0 +1,79 @@
...
+        assertThat("There should be 2 related mqtt server config parameters",
+                bridgeConfig.getMqttConfig().getConfig().size(), is(2));
+        assertThat("Mqtt server host should be 'localhost'",
+                bridgeConfig.getMqttConfig().getHost(), is("localhost"));
+        assertThat("Mqtt server port should be '1883'",
+                bridgeConfig.getMqttConfig().getPort(), is(1883));
+
+        // Kafka server config related tests
+        assertThat("There should be 2 related kafka config parameters",
+                bridgeConfig.getKafkaConfig().getConfig().size(), is(2));
```
kyguy, line 44: Was the application.properties file updated recently? 

This test was failing locally for me, just want to confirm it isn't just an env issue on my side

kyguy (approved):
Looks solid @antonio-pedro99, left some minor nits

src/test/java/io/strimzi/kafka/bridge/mqtt/core/ConfigRetrieverTest.java
antonio-pedro99: Yes, sorry! I forgot to update this test. 
There is only one configuration parameter for Kafka now. 
We have removed configuration for the producer.

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
antonio-pedro99: Marked as solved

src/test/java/io/strimzi/kafka/bridge/mqtt/core/ConfigRetrieverTest.java
antonio-pedro99: it is sorted now.

PR #12, opened 2023-06-30
Add producer

antonio-pedro99:
This is a draft PR for initial assumptions about the Kafka Producer.

### Design

`BridgeKafkaProducer`: represents a Kafka Producer. It has an internal producer client. 

`BridgeKafkaProducerFactory`: responsible for creating instantiate a `BridgeKafkaProducer` according to the QoS of the incoming MQTT Message. This entity helps to ensure that the bridges will have only two Producers, one for handling QoS 0 and the other QoS 1. 

The `BridgeKafkaProducerFactory` is also responsible for receiving the Kafka configuration from the upper level of the Bridge.


antonio-pedro99:
DW about unnecessary changes, I am going to rebase this branch.

src/main/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducer.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+/**
+ * Represents a Kafka producer for the Bridge.
+ */
+public class BridgeKafkaProducer<K, V> {
+
+
+    private final int ackLevel;
```
kyguy, line 20: Would it be better if we used an enum instead of int here?

We could even add some comments in the enum explaining the different ACK levels

Let me know what you think!

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -91,15 +95,57 @@ private void handleConnectMessage(ChannelHandlerContext ctx) {
...
      *
      * @param ctx            ChannelHandlerContext instance
      * @param publishMessage represents a MqttPublishMessage
      */
     private void handlePublishMessage(ChannelHandlerContext ctx, MqttPublishMessage publishMessage) {
-        logger.info("MQTT Topic: {}", publishMessage.variableHeader().topicName());
-        logger.info("Kafka Mapped Topic: {}", mqttKafkaMapper.map(publishMessage.variableHeader().topicName()));
-        logger.info("Message: {}", publishMessage.payload().toString(Charset.defaultCharset()));
+        // get QoS level from the MqttPublishMessage
+        int qos = publishMessage.fixedHeader().qosLevel().value();
```
kyguy, line 120: Should we use enum here?

src/main/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducer.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * Close the producer
+     */
+    public void close() {
+        if (this.clientProducer != null) {
+            this.clientProducer.close();
+        }
+    }
```
kyguy, line 69: I notice this method hasn't been used yet, should this be called by BridgeKafkaProducerFactory class when an exception with the bridge occurs?

kyguy (commented):
Left a couple questions/nits, looking good @antonio-pedro99

src/main/java/io/strimzi/kafka/bridge/mqtt/config/KafkaConfig.java
```diff
@@ -53,11 +53,19 @@ public KafkaProducerConfig getKafkaProducerConfig() {
         return kafkaProducerConfig;
     }
 
+
+    @Override
+    public Map<String, Object> getConfig() {
+        // get the kafka config without the prefix
+        return this.config.entrySet().stream().collect(Collectors.toMap((e)->e.getKey().substring(6), Map.Entry::getValue));
```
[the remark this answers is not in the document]
antonio-pedro99: Even if it is only used once?
Maybe a local constant in getConfig() ?

src/main/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducer.java
```diff
@@ -0,0 +1,70 @@
...
+    public void create(KafkaConfig kafkaConfig) {
+        Properties props = new Properties();
+        String keySerializer = StringSerializer.class.getName();
+        String valueSerializer = StringSerializer.class.getName();
+        props.putAll(kafkaConfig.getConfig());
+        props.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, keySerializer);
+        props.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, valueSerializer);
+        props.put(ProducerConfig.ACKS_CONFIG, String.valueOf(this.ackLevel));
+        this.clientProducer = new KafkaProducer<>(props);
+    }
```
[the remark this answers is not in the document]
antonio-pedro99: I was following the HTTP Bridge impl.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -91,15 +95,57 @@ private void handleConnectMessage(ChannelHandlerContext ctx) {
...
+
+        // build the Kafka record
+        ProducerRecord<String, Object> record = new ProducerRecord<>(mappedTopic,
+                publishMessage.payload().toString());
+
+        // get the appropriate Kafka producer according to the QoS level
+        BridgeKafkaProducer<String, Object> producer = BridgeKafkaProducerFactory.getInstance().getProducer(qos);
+
+        // send the record to the Kafka topic
+        if (qos == 0) {
```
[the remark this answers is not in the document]
antonio-pedro99: yes, that would make things clearer.

src/main/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducer.java
antonio-pedro99: yes, that would be better

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
antonio-pedro99: how would it look like here?

src/main/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducer.java
antonio-pedro99: Yes, that is the idea.

src/main/java/io/strimzi/kafka/bridge/mqtt/config/KafkaConfig.java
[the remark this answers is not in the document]
antonio-pedro99: or better use the KAFKA_CONFIG_PREFIX.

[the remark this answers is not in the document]
ppatierno: I don't see we have something similar in the Strimzi bridge. The `KafkaConfig` class doesn't override the `getConfig` from the `AbstractConfig`. It's worth to take a look at there.

src/main/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducer.java
ppatierno: I think this should be a `String` in order to be prepared for using `all` other than `0` and `1` as possible values.

src/main/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducerFactory.java
```diff
@@ -0,0 +1,72 @@
...
+package io.strimzi.kafka.bridge.mqtt.kafka;
+
+import io.strimzi.kafka.bridge.mqtt.config.KafkaConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Factory for creating Kafka producers as needed.
+ */
+public class BridgeKafkaProducerFactory<K, V> {
```
ppatierno, line 14: not sure how the factory approach applies here. With the Factory pattern you simplify the way to build object of a different shapes but implementing the same interface. In this case it would be more having a `BridgeKafkaProducer` interface and then two concrete classes `BridgeKafkaProducerZero` and `BridgeKafkaProducerOne` which doesn't need an `ack` parameter in the constructor. Then the factory could also have just static methods to build each of them.
Other than this, a factory should not take responsability to hold the reference to what it builds and being a singleton to do so for sharing the producers across the application.
Let's try to think a better approach here.

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
```diff
@@ -6,7 +6,7 @@
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.strimzi.kafka.bridge.mqtt.mapper.MappingRule;
-
+import io.strimzi.kafka.bridge.mqtt.mapper.MqttKafkaMapper;
```
ppatierno, line 9: why this additional import here?

src/main/java/io/strimzi/kafka/bridge/mqtt/config/KafkaConfig.java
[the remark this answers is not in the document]
antonio-pedro99: yes, you are right!
but if we call getConfig, we would get the prefix like kafka., and it was causing issue while creating the producer.

src/main/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducer.java
antonio-pedro99: "all" is equivalent to -1.

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
antonio-pedro99: This should not be here!

src/main/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducerFactory.java
antonio-pedro99: While the name suggests that it is a Factory, I would not say it is using exactly the same factory concept. I was just relying on some of its principle, and I came to this naive solution so far. 
To start, I was thinking of ways to get the specific producer (with ACK 1 or 0) without creating a producer every time a new MQTT message is received. 
That's why I came with this factory look-alike component, it just receives the config that will be shared across the producers, decide when to create the producers(only once), and the getProducer(QoS). 

Yes, I am considering the points you mentioned, and I will think about something better.

ppatierno: Also take into account, as we saw for the loader class, that testing a singleton is not that simple, so if possible I would avoid it.

> To start, I was thinking of ways to get the specific producer (with ACK 1 or 0) without creating a producer every time a new MQTT message is received.

Absolutely. We don't need one producer per MQTT client/message.

src/main/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducer.java
ppatierno: let's keep the semantic we have in Kafka, 0, 1, all ... not moving to something different. You never know if at some point Kafka will start using -1 with a specific meaning.

src/main/java/io/strimzi/kafka/bridge/mqtt/config/KafkaConfig.java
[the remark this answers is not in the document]
ppatierno: I should check the Strimzi bridge code, because it's working there.

src/main/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducerFactory.java
antonio-pedro99: I had a hard time testing singleton :(
I need to investigate more and come with a good solution.

src/main/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducer.java
antonio-pedro99: Anyway, I have changed this.
Now I am using enum as suggested by @kyguy .

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MqttMessageQoS.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.bridge.mqtt.utils;
+
+/**
+ * Represents the MQTT message QoS
+ */
+public enum MqttMessageQoS {
```
ppatierno, line 10: no need for this class, Netty library has its own you can use `io.netty.handler.codec.mqtt.MqttQoS`

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServer.java
```diff
@@ -29,41 +31,47 @@ public class MqttServer {
...
-                .childHandler(new MqttServer.MqttServerInitializer())
+                .childHandler(new MqttServer.MqttServerInitializer(config.getKafkaConfig()))
                 .childOption(option, true);
     }
 
+    /**
+     * This helper class help us add necessary Netty pipelines handlers. <br>
+     * During the {@link #initChannel(SocketChannel)}, we use MqttDecoder() and MqttEncoder to decode and encode Mqtt messages respectively. <br>
+     */
+    private static class MqttServerInitializer extends ChannelInitializer<SocketChannel> {
```
antonio-pedro99, line 60: should we move this to a separate file?

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -4,46 +4,62 @@
...
 import io.netty.channel.ChannelHandlerContext;
-import io.netty.channel.SimpleChannelInboundHandler;
+import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.handler.codec.mqtt.MqttMessage;
 import io.netty.handler.codec.mqtt.MqttMessageBuilders;
 import io.netty.handler.codec.mqtt.MqttMessageType;
 import io.netty.handler.codec.mqtt.MqttPublishMessage;
 import io.netty.handler.codec.mqtt.MqttConnectReturnCode;
 import io.netty.handler.codec.mqtt.MqttConnAckMessage;
+import io.netty.handler.codec.mqtt.MqttQoS;
```
antonio-pedro99, line 16: Sorry for the last PR, my IDE was rearranging imports automatically before any commit

src/main/java/io/strimzi/kafka/bridge/mqtt/config/KafkaConfig.java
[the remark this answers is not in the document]
antonio-pedro99: closed as fixed

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServer.java
kyguy: I think it might be neater if we move it to another file

kyguy (commented):
Looking great @antonio-pedro99, assuming we are leaving EXACTLY_ONCE for some later PR, I think all that we have left to do here is adding some unit tests for the new logic. This will involve mocking a Kafka broker and verifying the producer works as intended when communicating with it

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServer.java
ppatierno: I agree.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -91,15 +110,76 @@ private void handleConnectMessage(ChannelHandlerContext ctx) {
...
-        logger.info("Message: {}", publishMessage.payload().toString(Charset.defaultCharset()));
+        // get QoS level from the MqttPublishMessage
+        MqttQoS qos = MqttQoS.valueOf(publishMessage.fixedHeader().qosLevel().value());
+
+        // perform topic mapping
+        String mappedTopic = mqttKafkaMapper.map(publishMessage.variableHeader().topicName());
+
+        // build the Kafka record
+        ProducerRecord<String, ByteBuf> record = new ProducerRecord<>(mappedTopic,
+                publishMessage.payload());
```
ppatierno, line 142: We are just moving bytes from MQTT payload to Kafka payload, but it doesn't allow to the MQTT client to specify a key (see above comment).
Usually the MQTT payload is something more than just bytes, i.e. a JSON that could contains key and value as we have for the Strimzi HTTP bridge. The JSON could also bring more than one message/record.
As starting point I am fine with moving bytes and not having a key, but we should track with an issue and think about it as next step.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -91,15 +110,76 @@ private void handleConnectMessage(ChannelHandlerContext ctx) {
...
+                    if (error != null) {
+                        logger.error("Error sending message to Kafka: ", error);
+                    } else {
+                        logger.info("Message sent to Kafka on topic {} with offset {}", metadata.topic(), metadata.offset());
+                        // send PUBACK message to the client
+                        sendPubAckMessage(ctx, publishMessage.variableHeader().packetId());
+                    }
+                });
+            }
+            case EXACTLY_ONCE -> logger.warn("QoS level EXACTLY_ONCE is not supported yet");
```
ppatierno, line 166:
```suggestion
            case EXACTLY_ONCE -> logger.warn("QoS level EXACTLY_ONCE is not supported");
```

Let's not provide future expectation :-)

src/main/java/io/strimzi/kafka/bridge/mqtt/utils/ByteBufSerializer.java
```diff
@@ -0,0 +1,30 @@
...
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufInputStream;
+import org.apache.kafka.common.errors.SerializationException;
+import org.apache.kafka.common.serialization.Serializer;
+import java.io.IOException;
+
+/**
+ * Serializer for ByteBuf
+ */
+public class ByteBufSerializer implements Serializer<ByteBuf> {
```
ppatierno, line 16: See my previous comment, I don't think we need this serializer.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -91,15 +110,76 @@ private void handleConnectMessage(ChannelHandlerContext ctx) {
...
-        logger.info("Kafka Mapped Topic: {}", mqttKafkaMapper.map(publishMessage.variableHeader().topicName()));
-        logger.info("Message: {}", publishMessage.payload().toString(Charset.defaultCharset()));
+        // get QoS level from the MqttPublishMessage
+        MqttQoS qos = MqttQoS.valueOf(publishMessage.fixedHeader().qosLevel().value());
+
+        // perform topic mapping
+        String mappedTopic = mqttKafkaMapper.map(publishMessage.variableHeader().topicName());
+
+        // build the Kafka record
+        ProducerRecord<String, ByteBuf> record = new ProducerRecord<>(mappedTopic,
```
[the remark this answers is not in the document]
antonio-pedro99: We can not access the bytes directly from `ByteBuf`. 
To access it, we would need some workaround like this
```java
try {
            byte[] serializedData = new byte[data.readableBytes()];

            try (ByteBufInputStream inputStream = new ByteBufInputStream(data)) {
                inputStream.read(serializedData);
            }
            return serializedData;
        } catch (IOException e) {
            throw new SerializationException("Error serializing ByteBuf", e);
        }
```
And this is exactly how I implemented the `ByteBufSerializer`.
Is there any other way to get the bytes? I could not find it yet.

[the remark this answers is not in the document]
ppatierno: Isn't the `array()` method what we are looking for? https://netty.io/4.1/api/io/netty/buffer/ByteBuf.html#array--

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -73,6 +90,8 @@ protected void channelRead0(ChannelHandlerContext ctx, MqttMessage msg) {
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
         cause.printStackTrace();
+        this.bridgeKafkaProducerOne.close();
```
antonio-pedro99, line 93: I mistakenly closed the producer in a wrong callback. 
exceptionCaught(ChannelHandlerContext, Throwable) is invoked when an exception occurs inside a channel (client connection), we can not close the producers if a client caught an exception. 
so I have moved this to the appropriate callback, the one that is invoked when the handler is removed from the pipeline, meaning that it will no longer handler channels (because the server has topped for example)

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -91,15 +110,76 @@ private void handleConnectMessage(ChannelHandlerContext ctx) {
...
      */
     private void handlePublishMessage(ChannelHandlerContext ctx, MqttPublishMessage publishMessage) {
-        logger.info("MQTT Topic: {}", publishMessage.variableHeader().topicName());
-        logger.info("Kafka Mapped Topic: {}", mqttKafkaMapper.map(publishMessage.variableHeader().topicName()));
-        logger.info("Message: {}", publishMessage.payload().toString(Charset.defaultCharset()));
+        // get QoS level from the MqttPublishMessage
+        MqttQoS qos = MqttQoS.valueOf(publishMessage.fixedHeader().qosLevel().value());
+
+        // perform topic mapping
+        String mappedTopic = mqttKafkaMapper.map(publishMessage.variableHeader().topicName());
```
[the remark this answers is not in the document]
antonio-pedro99: sure.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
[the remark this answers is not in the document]
antonio-pedro99: my bad! I did not see it early.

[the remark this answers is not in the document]
antonio-pedro99: I would agree.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
antonio-pedro99, line 138: maybe `kafkaMappedTopic` or would make more sense?

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
antonio-pedro99: I have opened #20 to keep track of this.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
ppatierno: Agree

src/test/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducerTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+                    assertThat("Topic is correct",
+                            r.topic(), is(kafkaTopic));
+
+                    assertThat("Value is correct",
+                            r.value(), is("test".getBytes()));
+
+                    CompletableFuture<RecordMetadata> promise = new CompletableFuture<>();
+
+                    promise.complete(new RecordMetadata(new TopicPartition(r.topic(), 2), 234L, 0, 1000, 0L, 0, 0));
```
ppatierno, line 50:
```suggestion
                    promise.complete(new RecordMetadata(new TopicPartition(r.topic(), 2), 234L, 0, 1000, 0L, 0, "test".getBytes().size()));
```

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -91,15 +115,82 @@ private void handleConnectMessage(ChannelHandlerContext ctx) {
...
+
+        // perform topic mapping
+        String kafkaMappedTopic = mqttKafkaMapper.map(mqttTopic);
+
+        //log the topic mapping
+        logger.info("MQTT topic {} mapped to Kafka Topic {}", mqttTopic, kafkaMappedTopic);
+
+        // build the Kafka record
+        ProducerRecord<String, byte[]> record = new ProducerRecord<>(kafkaMappedTopic,
+                publishMessage.payload().array());
```
ppatierno, line 153: It seems not working, I got the following exception ...

```shell
[2023-07-09 15:32:09,537] ERROR <erverHandler:84> [oopGroup-3-1] Error reading Channel: 
java.lang.UnsupportedOperationException: direct buffer
	at io.netty.buffer.PooledUnsafeDirectByteBuf.array(PooledUnsafeDirectByteBuf.java:226) ~[netty-buffer-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.buffer.AbstractPooledDerivedByteBuf.array(AbstractPooledDerivedByteBuf.java:118) ~[netty-buffer-4.1.94.Final.jar:4.1.94.Final]
	at io.strimzi.kafka.bridge.mqtt.core.MqttServerHandler.handlePublishMessage(MqttServerHandler.java:153) ~[classes/:?]
	at io.strimzi.kafka.bridge.mqtt.core.MqttServerHandler.channelRead0(MqttServerHandler.java:80) ~[classes/:?]
	at io.strimzi.kafka.bridge.mqtt.core.MqttServerHandler.channelRead0(MqttServerHandler.java:39) ~[classes/:?]
```

Having a very quick look, it seems something like this is the way to go:

```java
byte[] data = new byte[publishMessage.payload().readableBytes()];
publishMessage.payload().readBytes(data);
// build the Kafka record
ProducerRecord<String, byte[]> record = new ProducerRecord<>(kafkaMappedTopic, data);
```

It's just worth checking if there is any better way.

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -76,6 +92,14 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
         ctx.close();
     }
 
+    @Override
+    public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
+        bridgeKafkaProducerWithNoAck.close();
+        bridgeKafkaProducerWithAckOne.close();
+        ctx.close();
+        super.handlerRemoved(ctx);
+    }
```
ppatierno, line 101: I think this has a problem, not sure how you tested it.
Try to open an MQTT client connection with client 1 and send messages, then DON'T close the connection.
Open another MQTT client connection with client 2 and send messages. Until now everything ok.
Close MQTT connection of client 1, it's going to close the Kafka producers and still connected MQTT client 2 won't be able to send messages anymore.

```shell
java.lang.IllegalStateException: Cannot perform operation after producer has been closed
	at org.apache.kafka.clients.producer.KafkaProducer.throwIfProducerClosed(KafkaProducer.java:969) ~[kafka-clients-3.5.0.jar:?]
	at org.apache.kafka.clients.producer.KafkaProducer.doSend(KafkaProducer.java:991) ~[kafka-clients-3.5.0.jar:?]
	at org.apache.kafka.clients.producer.KafkaProducer.send(KafkaProducer.java:962) ~[kafka-clients-3.5.0.jar:?]
	at io.strimzi.kafka.bridge.mqtt.kafka.BridgeKafkaProducer.send(BridgeKafkaProducer.java:45) ~[classes/:?]
	at io.strimzi.kafka.bridge.mqtt.core.MqttServerHandler.handlePublishMessage(MqttServerHandler.java:166) ~[classes/:?]
	at io.strimzi.kafka.bridge.mqtt.core.MqttServerHandler.channelRead0(MqttServerHandler.java:80) ~[classes/:?]
	at io.strimzi.kafka.bridge.mqtt.core.MqttServerHandler.channelRead0(MqttServerHandler.java:39) ~[classes/:?]
	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:99) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:346) ~[netty-codec-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:318) ~[netty-codec-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:440) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:919) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:166) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:788) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:724) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:650) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:562) ~[netty-transport-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:997) ~[netty-common-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) ~[netty-common-4.1.94.Final.jar:4.1.94.Final]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) ~[netty-common-4.1.94.Final.jar:4.1.94.Final]
	at java.lang.Thread.run(Thread.java:833) ~[?:?]
```

This is because the Kafka producers were closed when MQTT client 1 disconnected.

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
antonio-pedro99: @ppatierno what about using the previous approach?
It was actually doing the same as the code above, plus handling exceptions. I would go with a custom Serializer as it gives the possibility of having single responsibility, thus more readable and maintainable code. I mean, let us not add s many codes not related to the main task of handling an incoming message.

ppatierno: No we should stick with the default byte serializer. There is really no reason to have a new one doing exactly the same as the Kafka one just because of the Netty ByteBuf. Let's just get the bytes as Netty allows us.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
antonio-pedro99: I tested sequentially. First all clients connect, and randomly all of them start producing message to the server, as soon as a client is done producing, it should disconnect. This is kind same the below sequence. Maybe I missed some testing after implementing this callback. Please, allow me some time to investigate further.

antonio-pedro99: this callback should not be invoked on a client disconnection, according to the [netty docs ](https://netty.io/4.0/api/io/netty/channel/ChannelHandler.html#:~:text=void-,handlerRemoved(ChannelHandlerContext%C2%A0ctx),removed%20from%20the%20actual%20context%20and%20it%20doesn%27t%20handle%20events%20anymore.,-Method%20Detail), maybe I got it wrong!

antonio-pedro99: It is not working as expected :(.

I just connected all clients and without producing messages, I closed all of them, it this is making the producers to get closed. It should not work like! The producers just close in case the handler is destroyed or being closed.

antonio-pedro99: I believe there is no a way to do this inside the `SimpleChannelInboundHandler`.
Even when our current handler is sharable among channels, the [callbacks](https://netty.io/4.0/api/io/netty/channel/SimpleChannelInboundHandler.html#:~:text=Nested%20classes/interfaces%20inherited%20from%20interface%C2%A0io.netty.channel.ChannelHandler) have effects with respect to the current channel and not the handler itself. 

To be more clear, a very good [illustration](https://medium.com/geekculture/a-tour-of-netty-5020ecee5494#:~:text=each%20core%20component.-,Netty%20threading%20model%2C%20source%3A%20https%3A//blog.e%2Dkursy.it/netty%2Dintroduction/,-Netty%20ByteBuf) shows how channels are handled among event loops (worker groups). 

image

We have different pipelines for different channels (open connection), and each of them have a `ChannelHandler`, in our case, the `ChannelHandler` is `sharable`, so the same instance of it are being used among different channels, which means if we close the producer anywhere inside the `ChannelHandler`, we are going to close for all. 

Maybe we should consider providing global state and close the producers on the bridge shutdown?

antonio-pedro99:
I have added recent changes.

src/main/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducerService.java
```diff
@@ -4,54 +4,88 @@
...
 
+import java.util.Properties;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 
 /**
- * A service to handle the Kafka Producers
+ * Represents a Kafka producer for the Bridge.
  */
 public class BridgeKafkaProducerService {
```
ppatierno, line 24: Can we stick with what we have in the HTTP bridge for the name, it will be simpler to deal with comparisons. I would say it could be just `KafkaBridgeProducer`.

src/test/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducerServiceTest.java
```diff
@@ -13,29 +13,33 @@
...
-public class BridgeKafkaProducerTest {
+public class BridgeKafkaProducerServiceTest {
 
     /**
-     * Test the {@link BridgeKafkaProducer#send(ProducerRecord)} method
+     * Test the {@link BridgeKafkaProducerService#send(ProducerRecord)}} method
      */
     @Test
-    public void testProducerNoAck() {
+    public void testProducerSend() {
```
ppatierno, line 31: It's not totally clear how this test would work even because you are doing the sending with and without ACK together. It would be better to have two different tests to simplify the scenarios.

ppatierno (commented):
@antonio-pedro99 we are close, I left a couple of comments.

src/test/java/io/strimzi/kafka/bridge/mqtt/kafka/BridgeKafkaProducerServiceTest.java
antonio-pedro99: I was thinking about this.
But for noAck send, I do not think there are much to test.

```java
doNothing().when(producer).sendNoAck(any(ProducerRecord.class));
producer.sendNoAck(record);
```
and then what should be next?

kyguy (approved):
Great work @antonio-pedro99, as discussed on call we can add more tests later!

PR #13, opened 2023-06-30
Fixed imports with mapping related classes in the new mapper package

ppatierno:
The mapper loader was not using the classes from the new `mapper` package but still referring them from the old `utils`.
This PR fixes it.


ppatierno:
@antonio-pedro99 FYI

kyguy (approved):

PR #15, opened 2023-06-30
fix: testLoadRules

### LINKED ISSUE #14: The `testLoadRules` fails depending on the order of execution
The test `testLoadRules` in the `MappingRulesLoaderTest` class can fail depending on the order when it's executed.
If it runs first everything works fine, but if it runs after the `testInitMoreThanOnce`, it fails.
This is because the usage of singleton in `MappingRulesLoader` which is of course "shared" across tests.
If `testInitMoreThanOnce` runs first, it initialize the loader so the `testLoadRules` will fail.

antonio-pedro99:
This PR fixes #14 

Signed-off-by: Antonio Pedro <tonio.pedro99@gmail.com>

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
+++ b/src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
@@ -46,7 +46,7 @@ public class MappingRulesLoader {
      *
      * @return the singleton instance of the MappingRulesLoader
      */
-    public static MappingRulesLoader getInstance() throws NullPointerException {
+    public static synchronized MappingRulesLoader getInstance() throws NullPointerException {
         if (INSTANCE == null) {
             INSTANCE = new MappingRulesLoader();
         }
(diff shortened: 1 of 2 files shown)


src/main/java/io/strimzi/kafka/bridge/mqtt/utils/MappingRulesLoader.java
```diff
@@ -46,7 +46,10 @@ private MappingRulesLoader() {
      *
      * @return the singleton instance of the MappingRulesLoader
      */
-    public static MappingRulesLoader getInstance() {
+    public static MappingRulesLoader getInstance() throws NullPointerException {
```
AlexJuca, line 49: @antonio-pedro99 Should thread safety be a consideration here?

In its current implementation, multiple threads can potentially access the getInstance() method simultaneously and create multiple instances of the MappingRulesLoader class, which contradicts the intention of having a singleton pattern. If this is not a concern, kindly ignore this. 

For more info on the topic check out this article:

https://docs.oracle.com/javase/tutorial/essential/concurrency/sync.html

antonio-pedro99: Yeap, you are right!
Thanks for point it out @AlexJuca.

ppatierno (commented):
I don't think this is the right way to go for testing a singleton class. Just adding a close method for testing reveals an antipattern for singleton, when usually you do not need to close it across the application but just having one instance forever. 
The way to go for testing should be mocking the singleton for each test. You could use mockito to do that.

antonio-pedro99:
@ppatierno thanks for pointing down. 
I am on it.

antonio-pedro99:
Closed as the recent #17 fixes this issue.

PR #25, opened 2023-07-15
feat: added a new MqttKafkaMapper and updated docs.

### LINKED ISSUE #20: We are just moving bytes from MQTT payload to Kafka payload, but it doesn't allow to the MQTT client to specify a key (see above comment).
We are just moving bytes from MQTT payload to Kafka payload, but it doesn't allow to the MQTT client to specify a key (see above comment).
Usually the MQTT payload is something more than just bytes, i.e. a JSON that could contains key and value as we have for the Strimzi HTTP bridge. The JSON could also bring more than one message/record.
As starting point I am fine with moving bytes and not having a key, but we should track with an issue and think about it as next step.

_Originally posted by @ppatierno in https://github.com/strimzi/strimzi-mqtt-bridge/pull/12#discussion_r1256054715_

antonio-pedro99:
This PR includes:
- Added `kafkaKey` in the Topic Mapping Rules (this will fix #20).
- New way to define MQTT topic pattern: Now users can use regex as MQTT patterns, and gives the user more possibility to define rules.
- Placeholders are no longer defined as `{placeholder}`: Now we are using positional placeholders like $1, which means, get the value of the first capturing group in the MQTT pattern.
- The MqttKafkaMapper.map(String) does not return a String: Now the result of the mapping consists of a Kafka Topic and a Kafka record key, and they are wrapped in an object of a class called `MappingResult`.
-Tests for the new Mapper.
- Updated docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,8 +57,8 @@ The expression `([^/]+)` is used to represent the wildcard `+`, which in turn re
 It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
 
 Placeholders in the `KafkaTopic` and `kafkaKey` templates are defined using the `$` character followed by a number. 
-The number represents the index of the capturing group in the `mqttTopic` pattern.
-The MQTT Bridge uses the capturing groups in the `mqttTopic` pattern to positionally extract the values that will be used to replace the placeholders in the `KafkaTopic` and `kafkaKey` templates.
+The number represents the index of the `capturing group` in the `mqttTopic` pattern. `Please note that the index starts from 1 and not 0`.
+The MQTT Bridge uses the `capturing groups` in the `mqttTopic` pattern to positionally extract the values that will be used to replace the placeholders in the `KafkaTopic` and `kafkaKey` templates.
 
 Let's go through each rule in the above example to understand how the MQTT Bridge uses these rules to map MQTT topics to Kafka topics:
 
--- a/README.md
+++ b/README.md
@@ -36,23 +36,22 @@ The following is an example of a valid ToMaR:
 ```json
 [
   {
-    "mqttTopic": "building/(\\w+)/room/(\\d{1,4}).*",
-    "kafkaTopic": "building_$1_room_$2",
-    "kafkaKey": "room_$2"
+     "mqttTopic": "building/(\\w+)/room/(\\d{1,4}).*",
+     "kafkaTopic": "building_$1",
+     "kafkaKey": "room_$2"
   },
   {
     "mqttTopic": "sensors/([^/]+)/data",
     "kafkaTopic": "sensor_data"
   },
   {
-    "mqttTopic": "sensors.*",
-    "kafkaTopic": "sensor_others",
-    "kafkaKey": "sensor"
+     "mqttTopic": "sensors.*",
+     "kafkaTopic": "sensor_others"
   },
   {
     "mqttTopic": "devices/([^/]+)/data/(\b(all|new|old)\b)",
-    "kafkaTopic": "device_$1_data",
-    "kafkaKey": "device_$1"
+     "kafkaTopic": "device_$1_data",
+     "kafkaKey": "device_$2"
   }
 ]
 ```
@@ -73,32 +72,33 @@ used to replace the placeholders in the `KafkaTopic` and `kafkaKey` templates.
 Let's go through each rule in the above example to understand how the MQTT Bridge uses these rules to map MQTT topics to
 Kafka topics:
 
-1. MQTT Topic: `building/(\\w+)/room/(\\d{1,4}).*` -> Kafka Topic: `building_$1_room_$2`
+1. MQTT Topic: `building/(\\w+)/room/(\\d{1,4}).*` -> Kafka Topic: `building_$1` with Kafka Key: `room_$2`
 
    This rule maps MQTT topics of the form `building/{some word}/room/{some number with 4 digits}/#` to the Kafka
-   topic `building_$1_room_$2`.
+   topic `building_$1`.
    For example, if the MQTT topic is `building/A/room/1003/floor/2` it will be mapped to the Kafka
-   topic `building_A_room_1003` with the key `room_1003`.
+   topic `building_A` with the key `room_1003`.
 
-2. MQTT Topic: `sensors/([^/]+)/data` -> Kafka Topic: `sensor_data`
+2. MQTT Topic: `sensors/([^/]+)/data` -> Kafka Topic: `sensor_data` with Kafka Key: `null`
 
    This rule maps MQTT topics of the form `sensors/+/data` to the Kafka topic `sensor_data`.
    For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data`.
    Because the `KafkaKey` is not defined, the key of the Kafka record will be `null`.
 
-3. MQTT Topic: `sensors.*` -> Kafka Topic: `sensor_others`
+3. MQTT Topic: `sensors.*` -> Kafka Topic: `sensor_others` with Kafka Key: `null`
 
    This rule maps any MQTT topic starting with `sensors/` followed by any number of levels in the hierarchy to the Kafka
    topic `sensor_others`.
    For example, if the MQTT topic is `sensors/temperature/living-room`, it will be mapped to the Kafka
    topic `sensor_others` with the key `sensor`.
+   Because the `KafkaKey` is not defined, the key of the Kafka record will be `null`.
 
-4. MQTT Topic: `devices/([^/]+)/data/(\b(all|new|old)\b)` -> Kafka Topic: `device_$1_data`
+4. MQTT Topic: `devices/([^/]+)/data/(\b(all|new|old)\b)` -> Kafka Topic: `device_$1_data` with Kafka Key: `device_$2`
 
    This rule maps MQTT topics of the form `devices/{some word}/data/{either all, new or old}` to the Kafka
    topic `device_$1_data`.
    For example, if the MQTT topic is `devices/thermostat/data/all`, it will be mapped to the Kafka
-   topic `device_thermostat_data` with the key `device_thermostat`.
+   topic `device_thermostat_data` with the key `device_all`.
    This example also shows the advantage of using regex in the `mqttTopic` pattern. The last capturing group in
    the `mqttTopic` should be a word boundary `\b` followed by either `all`, `new` or `old` and anything else will not
    match the pattern.
--- a/README.md
+++ b/README.md
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
 
 ### Topic Mapping Rules (ToMaR)
 
-The ToMaR is a set of patterns the user provides defining how the MQTT Bridge maps MQTT topic names to Kafka topic names.
-A Mapping Rule is a model that contains an `MQTT topic pattern` and a `Kafka topic template`. 
-All the incoming MQTT message's topic should match an `MQTT topic pattern` in the ToMaR so that the bridge knows in which Kafka topic to produce this message.
+The ToMaR is a set of patterns the user provides defining how the MQTT Bridge maps MQTT topic names to Kafka topic
+names.
+A Mapping Rule is a model that contains an `MQTT topic pattern`, a `Kafka topic template`, and optionally
+a `Kafka record key`.
+All the incoming MQTT message's topic should match an `MQTT topic pattern` in the ToMaR so that the bridge knows in
+which Kafka topic to produce this message.
 This Kafka topic is defined by a template, `Kafka Topic template`.
-However, if the incoming MQTT message's topic does not match any pattern in the ToMaR, the Bridge has a default Kafka topic where the incoming message will be mapped to, known as `messages_default`.
+However, if the incoming MQTT message's topic does not match any pattern in the ToMaR, the Bridge has a default Kafka
+topic where the incoming message will be mapped to, known as `messages_default`.
 
-A valid ToMaR is a JSON file that contains an array of mapping rules. 
+The optional `Kafka record key` is used to define the key of the Kafka record that will be produced to the Kafka topic.
+It is defined by a template as well, and its default value is `null`.
+A valid ToMaR is a JSON file that contains an array of mapping rules.
 Each mapping rule is a JSON object that contains two properties: `mqttTopic` and `kafkaTopic`.
-The following is an example of a valid TOMAR:
+The following is an example of a valid ToMaR:
 
 ```json
 [
   {
-    "mqttTopic": "building/{building}/room/{room}/#",
-    "kafkaTopic": "building_{building}_room_{room}"
+    "mqttTopic": "building/(\\w+)/room/(\\d{1,4}).*",
+    "kafkaTopic": "building_$1_room_$2",
+    "kafkaKey": "room_$2"
   },
   {
-    "mqttTopic": "sensors/+/data",
+    "mqttTopic": "sensors/([^/]+)/data",
     "kafkaTopic": "sensor_data"
   },
   {
-    "mqttTopic": "sensors/#",
-    "kafkaTopic": "sensor_others"
+    "mqttTopic": "sensors.*",
+    "kafkaTopic": "sensor_others",
+    "kafkaKey": "sensor"
+  },
+  {
+    "mqttTopic": "devices/([^/]+)/data/(\b(all|new|old)\b)",
+    "kafkaTopic": "device_$1_data",
+    "kafkaKey": "device_$1"
   }
 ]
 ```
 
-The wildcard `#` represents one or more levels in the MQTT topic hierarchy.
-The wildcard `+` represents a single level in the MQTT topic hierarchy.
-It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
+The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic
+hierarchy.
+The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT
+topic hierarchy.
+It worth's mentioning that it is the user's responsibility to adhere to
+the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106)
+when defining the MQTT topic patterns.
 
-The parts with curly braces in the MQTT topic pattern are called `placeholders`.
-The MQTT Bridge will extract the values of these placeholders from the MQTT topic and use them to construct the Kafka topic name using the `kafkaTopic` template.
+Placeholders in the `KafkaTopic` and `kafkaKey` template are defined using the `$` character followed by a number. The
+number represents the index of the capturing group in the `mqttTopic` pattern.
+The MQTT Bridge uses the capturing groups in the `mqttTopic` pattern to positionally extract the values that will be
+used to replace the placeholders in the `KafkaTopic` and `kafkaKey` templates.
 
-Let's go through each rule in the above example to understand how the MQTT Bridge uses these rules to map MQTT topics to Kafka topics:
+Let's go through each rule in the above example to understand how the MQTT Bridge uses these rules to map MQTT topics to
+Kafka topics:
 
-1. MQTT Topic: `building/{building}/room/{room}/#` -> Kafka Topic: `building_{building}_room_{room}`
+1. MQTT Topic: `building/(\\w+)/room/(\\d{1,4}).*` -> Kafka Topic: `building_$1_room_$2`
 
-    This rule maps MQTT topics of the form `building/{building}/room/{room}/#` to the Kafka topic `building_{building}_room_{room}`. 
-    For example, if the MQTT topic is `building/A/room/1/floor/2` it will be mapped to the Kafka topic `building_A_room_1_floor_2`.
+   This rule maps MQTT topics of the form `building/{some word}/room/{some number with 4 digits}/#` to the Kafka
+   topic `building_$1_room_$2`.
+   For example, if the MQTT topic is `building/A/room/1003/floor/2` it will be mapped to the Kafka
+   topic `building_A_room_1003` with the key `room_1003`.
 
-2. MQTT Topic: `sensors/+/data` -> Kafka Topic: `sensor_data`
+2. MQTT Topic: `sensors/([^/]+)/data` -> Kafka Topic: `sensor_data`
 
-    This rule maps MQTT topics of the form `sensors/+/data` to the Kafka topic `sensor_data`. 
-    For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data`.
+   This rule maps MQTT topics of the form `sensors/+/data` to the Kafka topic `sensor_data`.
+   For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data`.
+   Because the `KafkaKey` is not defined, the key of the Kafka record will be `null`.
 
-3. MQTT Topic: `sensors/#` -> Kafka Topic: `sensor_others`
+3. MQTT Topic: `sensors.*` -> Kafka Topic: `sensor_others`
 
-   This rule maps any MQTT topic starting with `sensors/` followed by any number of levels in the hierarchy to the Kafka topic `sensor_others`.
-   For example, if the MQTT topic is `sensors/temperature/living-room`,  it will be mapped to the Kafka topic `sensor_others`.
+   This rule maps any MQTT topic starting with `sensors/` followed by any number of levels in the hierarchy to the Kafka
+   topic `sensor_others`.
+   For example, if the MQTT topic is `sensors/temperature/living-room`, it will be mapped to the Kafka
+   topic `sensor_others` with the key `sensor`.
 
+4. MQTT Topic: `devices/([^/]+)/data/(\b(all|new|old)\b)` -> Kafka Topic: `device_$1_data`
+
+   This rule maps MQTT topics of the form `devices/{some word}/data/{either all, new or old}` to the Kafka
+   topic `device_$1_data`.
+   For example, if the MQTT topic is `devices/thermostat/data/all`, it will be mapped to the Kafka
+   topic `device_thermostat_data` with the key `device_thermostat`.
+   This example also shows the advantage of using regex in the `mqttTopic` pattern. The last capturing group in
+   the `mqttTopic` should be a word boundary `\b` followed by either `all`, `new` or `old` and anything else will not
+   match the pattern.
 
 The order in which the rules are defined is important.
 The MQTT Bridge will use the first rule that matches the MQTT topic.
-For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data` because `sensors/+/data` matches the MQTT topic before `sensors/#`. 
-If we swap the positions of the rules, the MQTT Bridge would use the `sensors/#` to map the MQTT topic to the Kafka topic `sensor_others`.
+For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data`
+because `sensors/([^/]+)/data` matches the MQTT topic before `sensors/#`.
+If we swap the positions of the rules, the MQTT Bridge would use the `sensors.*` to map the MQTT topic to the Kafka
+topic `sensor_others`.
 
 ### MQTT Bridge Configuration
 
--- a/README.md
+++ b/README.md
@@ -67,7 +67,10 @@ Therefore, you should note the following:
   On the other hand, the pattern `building/.*` will match `building/room`, `building/floor/room`, and so on, but not `buildingroom` and `building/`. 
   - You can also use the `(?:\\/.*)?$` wildcard to match the whole subtopic level of the pattern. 
   For example, the pattern `locations/([^/]+)(?:\\/.*)?$` will match `locations/city/luanda/angola`, `locations/city`, `locations/city/`, and so on. 
-  It's literally equivalent to `locations/+/#`.
+  It's literally equivalent to `locations/+/#`.  
+  Please note that the `(?:\\/.*)?$` wildcard is a non-capturing group, which means that it will not be used to replace the placeholders in the `kafkaTopic` and `kafkaKey` templates.
+  - The `(?:\\/.*)?$` wildcard is different from the `.*` wildcard. For example, the pattern `sensors.*` will match everything after `sensors`, seperated with a slash or not. For example, `sensors`, `sensors/`, `sensorsdata`, and so on. 
+  On the other hand, the pattern `sensors(?:\\/.*)?$` will match `sensors`, `sensors/`, `sensors/data`, and so on, but not `sensorsdata`.
 - The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT topic hierarchy.
 It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
 
--- a/README.md
+++ b/README.md
@@ -60,13 +60,13 @@ The following is an example of a valid ToMaR:
 The expressions `.*` and `(?:\\/.*)?$` are used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic hierarchy.
 However, there are some cases that can lead to unexpected behavior when using the these wildcards interchangeably . 
 Therefore, you should note the following:
-  - You can not use the `.*` wildcard in capturing groups. 
+  - You cannot use the `.*` wildcard in capturing groups. 
   For example, the pattern `building/(\\w+)/room/(\\d{1,4})/(.*)` is invalid because it will capture the whole subtopic of the pattern, and `$3` placeholder might include `/` characters, which will lead to an invalid Kafka topic name.
   - You can use the `.*` wildcard without a preceding `/` character, but it is not equivalent to when it is preceded by `/`. 
   For example, the pattern `building.*` will match `building`, `building/room`, `buildingroom`, and so on. 
   On the other hand, the pattern `building/.*` will match `building/room`, `building/floor/room`, and so on, but not `buildingroom` and `building/`. 
   - You can also use the `(?:\\/.*)?$` wildcard to match the whole subtopic level of the pattern. 
-  For example, the pattern `locations/([^/]+)(?:\\/.*)?$` will match `locations/city/luanda/angola`, `locations/country`, and so on. 
+  For example, the pattern `locations/([^/]+)(?:\\/.*)?$` will match `locations/city/luanda/angola`, `locations/city`, `locations/city/`, and so on. 
   It's literally equivalent to `locations/+/#`.
 - The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT topic hierarchy.
 It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Let's go through each rule in the above example to understand how the MQTT Bridg
 
    This rule maps MQTT topics of the form `sensors/+/data` to the Kafka topic `sensor_data`.
    For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data`.
-   Because the `KafkaKey` is not defined, the key of the Kafka record will be `null`.
+   Because the `kafkaKey` is not defined, the key of the Kafka record will be `null`.
 
 3. MQTT Topic: `sensors.*` -> Kafka Topic: `sensor_others` with Kafka Key: `null`
 
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
 
 ### Topic Mapping Rules (ToMaR)
 
-The ToMaR is a set of patterns the user provides defining how the MQTT Bridge maps MQTT topic names to Kafka topic
+The ToMaR is a set of patterns the user provides defining how the MQTT Bridge maps MQTT topic names to Kafka topic names.
 names.
 A Mapping Rule is a model that contains an `MQTT topic pattern`, a `Kafka topic template`, and optionally
 a `Kafka record key`.
--- a/README.md
+++ b/README.md
@@ -48,18 +48,26 @@ The following is an example of a valid ToMaR:
     "mqttTopic": "devices/([^/]+)/data/(\b(all|new|old)\b)",
      "kafkaTopic": "device_$1_data",
      "kafkaKey": "device_$2"
+  },
+  {
+      "mqttTopic": "locations/([^/]+)(?:\\/.*)?$",
+      "kafkaTopic": "locations",
+      "kafkaKey": "location_$1"
   }
 ]
 ```
 
-- The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic hierarchy.
-However, there are some cases that can lead to unexpected behavior when using the `.*` wildcard.
+The expressions `.*` and `(?:\\/.*)?$` are used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic hierarchy.
+However, there are some cases that can lead to unexpected behavior when using the these wildcards interchangeably . 
 Therefore, you should note the following:
-  - You cannot use the `.*` wildcard in capturing groups. 
+  - You can not use the `.*` wildcard in capturing groups. 
   For example, the pattern `building/(\\w+)/room/(\\d{1,4})/(.*)` is invalid because it will capture the whole subtopic of the pattern, and `$3` placeholder might include `/` characters, which will lead to an invalid Kafka topic name.
   - You can use the `.*` wildcard without a preceding `/` character, but it is not equivalent to when it is preceded by `/`. 
   For example, the pattern `building.*` will match `building`, `building/room`, `buildingroom`, and so on. 
-  On the other hand, the pattern `building/.*` will match `building`, `building/room`, `building/floor/room`, and so on, but not `buildingroom`. 
+  On the other hand, the pattern `building/.*` will match `building/room`, `building/floor/room`, and so on, but not `buildingroom` and `building/`. 
+  - You can also use the `(?:\\/.*)?$` wildcard to match the whole subtopic level of the pattern. 
+  For example, the pattern `locations/([^/]+)(?:\\/.*)?$` will match `locations/city/luanda/angola`, `locations/country`, and so on. 
+  It's literally equivalent to `locations/+/#`.
 - The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT topic hierarchy.
 It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
 
--- a/README.md
+++ b/README.md
@@ -56,9 +56,9 @@ The expression `.*` is used to represent the wildcard `#`, which in turn represe
 The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT topic hierarchy.
 It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
 
-Placeholders in the `KafkaTopic` and `kafkaKey` templates are defined using the `$` character followed by a number. 
+Placeholders in the `kafkaTopic` and `kafkaKey` templates are defined using the `$` character followed by a number. 
 The number represents the index of the `capturing group` in the `mqttTopic` pattern. `Please note that the index starts from 1 and not 0`.
-The MQTT Bridge uses the `capturing groups` in the `mqttTopic` pattern to positionally extract the values that will be used to replace the placeholders in the `KafkaTopic` and `kafkaKey` templates.
+The MQTT Bridge uses the `capturing groups` in the `mqttTopic` pattern to positionally extract the values that will be used to replace the placeholders in the `kafkaTopic` and `kafkaKey` templates.
 
 Let's go through each rule in the above example to understand how the MQTT Bridge uses these rules to map MQTT topics to Kafka topics:
 
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ The following is an example of a valid ToMaR:
 ```json
 [
   {
-     "mqttTopic": "building/(\\w+)/room/(\\d{1,4}).*",
+     "mqttTopic": "building/(\\w+)/room/(\\d{1,4})/.*",
      "kafkaTopic": "building_$1",
      "kafkaKey": "room_$2"
   },
@@ -52,8 +52,11 @@ The following is an example of a valid ToMaR:
 ]
 ```
 
-The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic hierarchy.
-The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT topic hierarchy.
+- The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic hierarchy.
+However, there are some cases that can lead to unexpected behavior when using the `.*` wildcard. Therefore, you should note the following:
+  - You can not use the `.*` wildcard in capturing groups. For example, the pattern `building/(\\w+)/room/(\\d{1,4})/(.*)` is invalid because it will capture the whole subtopic of the pattern, and `$3` placeholder might include `/` characters, which will lead to an invalid Kafka topic name.
+  - You can use the `.*` wildcard without a preceding `/` character, but it is not equivalent to when it is preceded by `/`. For example, the pattern `building.*` will match `building`, `building/room`, `buildingroom`, and so on. On the other hand, the pattern `building/.*` will match `building`, `building/room`, `building/floor/room`, and so on, but not `buildingroom`. 
+- The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT topic hierarchy.
 It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
 
 Placeholders in the `kafkaTopic` and `kafkaKey` templates are defined using the `$` character followed by a number. 
--- a/README.md
+++ b/README.md
@@ -18,14 +18,10 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
 ### Topic Mapping Rules (ToMaR)
 
 The ToMaR is a set of patterns the user provides defining how the MQTT Bridge maps MQTT topic names to Kafka topic names.
-names.
-A Mapping Rule is a model that contains an `MQTT topic pattern`, a `Kafka topic template`, and optionally
-a `Kafka record key`.
-All the incoming MQTT message's topic should match an `MQTT topic pattern` in the ToMaR so that the bridge knows in
-which Kafka topic to produce this message.
+A Mapping Rule is a model that contains an `MQTT topic pattern`, a `Kafka topic template`, and optionally a `Kafka record key`.
+All the incoming MQTT message's topic should match an `MQTT topic pattern` in the ToMaR so that the bridge knows in which Kafka topic to produce this message.
 This Kafka topic is defined by a template, `Kafka Topic template`.
-However, if the incoming MQTT message's topic does not match any pattern in the ToMaR, the Bridge has a default Kafka
-topic where the incoming message will be mapped to, known as `messages_default`.
+However, if the incoming MQTT message's topic does not match any pattern in the ToMaR, the Bridge has a default Kafka topic where the incoming message will be mapped to, known as `messages_default`.
 
 The optional `Kafka record key` is used to define the key of the Kafka record that will be produced to the Kafka topic.
 It is defined by a template as well, and its default value is `null`.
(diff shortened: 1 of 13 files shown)


src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -138,14 +139,17 @@ private void handlePublishMessage(ChannelHandlerContext ctx, MqttPublishMessage
         String mqttTopic = publishMessage.variableHeader().topicName();
 
         // perform topic mapping
-        String kafkaMappedTopic = mqttKafkaMapper.map(mqttTopic);
+        MappingResult mqttKafkaMappingResult = mqttKafkaMapper.map(mqttTopic);
+
+        String kafkaMappedTopic = mqttKafkaMappingResult.getKafkaTopic();
+        String key = mqttKafkaMappingResult.getKafkaKey();
```
ppatierno, line 145: we don't need to save the getter results in two local variables. The Java compiled will do optimization for you. Just call the getter more times when needed.

src/main/java/io/strimzi/kafka/bridge/mqtt/mapper/MappingResult.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+package io.strimzi.kafka.bridge.mqtt.mapper;
+
+/**
+ * Represents the result of a mapping operation.
+ * It contains the mapped kafka topic and the kafka key.
+ */
+public class MappingResult {
+    private String kafkaTopic;
+    private String kafkaKey;
```
ppatierno, line 13: I would make these two final so that they can be inizialized just in the constructor and removing the setters from this class.

src/main/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaMapper.java
```diff
@@ -46,113 +33,61 @@ public class MqttKafkaMapper {
...
-                            mappedKafkaTopic = mappedKafkaTopic.replace(entry.getKey(), entry.getValue());
-                        } else {
-                            throw new IllegalArgumentException("The placeholder " + entry.getKey() + " was not found assigned any value.");
-                        }
-                    }
+                if (kafkaKey != null) {
+                    // check for pending placeholders replacement in the kafka key.
+                    checkPlaceholder(kafkaKey);
+                    // set the mapped kafka key to the result
+                    mappingResult.setKafkaKey(kafkaKey);
```
ppatierno, line 72: related to the other comment in the `MappingResult` class, can we just create the `MappingResult` passing topic and key in the constructor instead of using setters in two places. Of course, key could be also null but it would be a valid parameter to the constructor.

src/test/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaMapperTest.java
```diff
@@ -42,28 +48,56 @@ public void testDefaultTopic() {
...
 
-        rules.add(new MappingRule("sensors/+/data", "sensor_data"));
-        rules.add(new MappingRule("devices/{device}/data", "devices_{device}_data"));
-        rules.add(new MappingRule("fleet/{fleet}/vehicle/{vehicle}", "fleet_{fleet}"));
-        rules.add(new MappingRule("building/{building}/floor/{floor}", "building.{building}.floor.{floor}"));
-        rules.add(new MappingRule("term/{number}", "term{number}"));
+        rules.add(new MappingRule("sensors/[^/]+/data", "sensor_data", null));
+        rules.add(new MappingRule("devices/([^/]+)/data", "devices_$1_data", "device_$1"));
+        rules.add(new MappingRule("fleet/([0-9]+)/vehicle/(\\w+)", "fleet_$1", "fleet$2"));
+        rules.add(new MappingRule("building/(\\d+)/floor/(\\d+).*", "building_$2_$1", "building_$1"));
```
ppatierno, line 54: can we make this example test clearer, just using the placeholders in the "logical" order $1_$2? 
Then you can have another test case where you have a different order, to show that it's possible.

src/main/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaMapper.java
```diff
@@ -46,113 +33,61 @@ public class MqttKafkaMapper {
...
      *
      * @param mqttTopic represents an MQTT topic.
-     * @return a valid Kafka topic.
+     * @return a MappingResult object containing the mapped kafka topic and kafka key.
      * @see MappingRule
+     * @see MappingResult
      */
-    public String map(String mqttTopic) {
+    public MappingResult map(String mqttTopic) {
+        MappingResult mappingResult = new MappingResult(DEFAULT_KAFKA_TOPIC, null);
```
ppatierno, line 50: You are creating an instance here and modifying it later with the setter.
See my other comments, I would create the instance only at the end when you figure out which topic and key to use. It will be clearer in the code where it's created and with which values instead of having to figure out when setters are called.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
antonio-pedro99: Maybe for more readability?

ppatierno: Not sure it's really more readable. But I can understand coming from university they teach it this way ;-)

antonio-pedro99: Calling mqttKafkaMappingResult.getKafkaTopic() all the time is kind tiresome for me. 
But if you prefer, I can use mqttKafkaMapper.map(mqttTopic).getKafkaTopic() every I need it.

src/main/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaMapper.java
antonio-pedro99: I just converted the MappingResult to a record, is that fine? We anyway do not need the setters.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
ppatierno: Nope this is wrong, because you are calling the map each time!
I meant calling the map, getting a MappingResult and then calling on it the getters multiple times instead of storing results in variables.

src/main/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaMapper.java
ppatierno: Yeah that makes sense.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
antonio-pedro99: Well we are just using kafkaMappedTopic and key two times then I think do not need to store. Thanks!

kyguy (commented):
Looking good @antonio-pedro99, I don't have anything to add on top of Paolo's comments here. I'll review again once  `MqttKafkaMapper` is abstracted for a `MqttKafkaSimpleMapper` class  with the original mapper logic and `MqttKafkaRegexMapper` class for the new mapper logic as discussed on Slack

src/main/java/io/strimzi/kafka/bridge/mqtt/mapper/MappingResult.java
antonio-pedro99: fixed sensei.

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -115,15 +127,6 @@ private void sendPubAckMessage(ChannelHandlerContext ctx, int packetId) {
         ctx.writeAndFlush(pubAckMessage);
     }
 
-    /**
```
antonio-pedro99, line 118: IntelliJ work :(

src/test/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaRegexMapperTest.java
```diff
@@ -0,0 +1,245 @@
...
+                mapper.map("fleet/4/vehicle/23").kafkaTopic(), is("fleet_4"));
+
+        assertThat("The key for fleet/([0-9]+)/vehicle/(\\w+) should be fleet$2",
+                mapper.map("fleet/4/vehicle/23").kafkaKey(), is("fleet23"));
+
+        assertThat("term/(\\d+) should be mapped to term$1",
+                mapper.map("term/4").kafkaTopic(), is("term4"));
+
+        assertThat("The key for term/(\\d+) should be my_term_key_$1",
+                mapper.map("term/4").kafkaKey(), is("my_term_key_4"));
```
ppatierno, line 88: For all the above you did what I already mentioned before. You should call `map` one time (for each pattern) and the run the asserts on the `MappingResult` instance.

src/test/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaRegexMapperTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+        assertThat("Mqtt pattern sport/tennis/(\\w+).* should be mapped to sports_$1",
+                mapper.map("sport/tennis/player123/score/wimbledon").kafkaTopic(), is("sports_player123"));
+
+        // Test for ([^/]+)/recipes.* pattern
+        assertThat("Mqtt pattern ([^/]+)/recipes.* should be mapped to my_recipes",
+                mapper.map("angolan/recipes/caculu/fish").kafkaTopic(), is("angolan_recipes"));
+
+        assertThat("The key for ([^/]+)/recipes.* should be expanded to $1",
+                mapper.map("angolan/recipes/caculu/fish").kafkaKey(), is("angolan"));
```
ppatierno, line 196: Usual note. Calling `map` once for each pattern and verify topic and key on the resulting `MappingResult`.

ppatierno (changes requested):
I left some comments but I would be more realistic about example rules and ones used in the tests about topic and key.
In general, does not make sense having the same placeholder used in both topic template and key template.
For example ...

mqtt = `building/([0-9]+)/room/([0-9]+)` 
topic = `building_$1_room_$2
key = `building_$1`

This way you are actually assigning to each building-room couple a dedicated topic and the key doesn't help at all. Let me explain why ...
Assume the client sends with building=10, room=2 ... topic is `building_10_room_2` and key `building_10`. Each message sent to this topic will have same key. If this topic has 5 partitions .... one will be filled (all messages same `building_10` key) and others will be just empty :-D
The same will apply for any other topic. I.e. building = 3, room = 5 ... topic is `building_3_room_5` using always key `building_3`.
The above mapping has any sense only if one partition but in general doesn't not make any sense.

A valid usage would be:

mqtt = `building/([0-9]+)/room/([0-9]+)` 
topic = `building_$1
key = `room_$2`

So building=10, room=2 message will go to topic `building_10` and key `room_2`, then messages with building=10, room=4 will got still to topic `building_10` and key `room_4`.
You are dedicating a topic per building and distributed messages across partition based on room. This makes totally sense :-)

src/main/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaRegexMapper.java
```diff
@@ -0,0 +1,79 @@
...
+                // return the first match
+                return new MappingResult(mappedKafkaTopic, kafkaKey);
+            }
+        }
+        return new MappingResult(MqttKafkaMapper.DEFAULT_KAFKA_TOPIC, null);
+    }
+
+    @Override
+    protected void buildOrCompilePatterns() {
+        this.rules.forEach(e -> patterns.add(Pattern.compile(e.getMqttTopicPattern())));
```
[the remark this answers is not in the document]
antonio-pedro99: in forEach context means `element` :-)

src/test/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaRegexMapperTest.java
```diff
@@ -0,0 +1,245 @@
...
+    public void testDefaultTopic() {
+        List<MappingRule> rules = new ArrayList<>();
+
+        MqttKafkaRegexMapper mapper = new MqttKafkaRegexMapper(rules);
+
+        assertThat("Should use the default topic when no mapping pattern matches.",
+                mapper.map("sensor/temperature").kafkaTopic(), is(MqttKafkaSimpleMapper.DEFAULT_KAFKA_TOPIC));
+
+        assertThat("The key for the default topic should be null",
+                mapper.map("sensor/temperature").kafkaKey(), nullValue());
```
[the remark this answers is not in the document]
antonio-pedro99: I was actually doing this, see the first commit. 
but as the number of tests increases, to avoid a lot of lines, I decided to call it when we need to test.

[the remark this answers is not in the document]
ppatierno: Yeah but did you get the current is somehow wrong? You are doing testing on two different instances of MappingResult because you call map two times. Back to the initial tests then. More lines but correct ones :-)

antonio-pedro99:
Thanks for the explanation, it makes total sense.
The same it would happen if we had for example: 
mqtt = building/#
topic = building_data
key = building
In this case, it would be better to use null, which means use round Robbin other than forcing everything in one partition because of the key `building`.

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
 
 ### Topic Mapping Rules (ToMaR)
 
-The ToMaR is a set of patterns the user provides defining how the MQTT Bridge maps MQTT topic names to Kafka topic names.
-A Mapping Rule is a model that contains an `MQTT topic pattern` and a `Kafka topic template`. 
-All the incoming MQTT message's topic should match an `MQTT topic pattern` in the ToMaR so that the bridge knows in which Kafka topic to produce this message.
+The ToMaR is a set of patterns the user provides defining how the MQTT Bridge maps MQTT topic names to Kafka topic
```
ppatierno, line 20:
```suggestion
The ToMaR is a set of patterns the user provides defining how the MQTT Bridge maps MQTT topic names to Kafka topic names.
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+names.
```
ppatierno, line 21:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+A Mapping Rule is a model that contains an `MQTT topic pattern`, a `Kafka topic template`, and optionally
+a `Kafka record key`.
```
ppatierno, line 23:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
 
 ### Topic Mapping Rules (ToMaR)
 
-The ToMaR is a set of patterns the user provides defining how the MQTT Bridge maps MQTT topic names to Kafka topic names.
-A Mapping Rule is a model that contains an `MQTT topic pattern` and a `Kafka topic template`. 
-All the incoming MQTT message's topic should match an `MQTT topic pattern` in the ToMaR so that the bridge knows in which Kafka topic to produce this message.
+The ToMaR is a set of patterns the user provides defining how the MQTT Bridge maps MQTT topic names to Kafka topic
+names.
+A Mapping Rule is a model that contains an `MQTT topic pattern`, a `Kafka topic template`, and optionally
```
ppatierno, line 22:
```suggestion
A Mapping Rule is a model that contains an `MQTT topic pattern`, a `Kafka topic template`, and optionally a `Kafka record key`.
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+a `Kafka record key`.
+All the incoming MQTT message's topic should match an `MQTT topic pattern` in the ToMaR so that the bridge knows in
+which Kafka topic to produce this message.
```
ppatierno, line 25:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
 ### Topic Mapping Rules (ToMaR)
 
-The ToMaR is a set of patterns the user provides defining how the MQTT Bridge maps MQTT topic names to Kafka topic names.
-A Mapping Rule is a model that contains an `MQTT topic pattern` and a `Kafka topic template`. 
-All the incoming MQTT message's topic should match an `MQTT topic pattern` in the ToMaR so that the bridge knows in which Kafka topic to produce this message.
+The ToMaR is a set of patterns the user provides defining how the MQTT Bridge maps MQTT topic names to Kafka topic
+names.
+A Mapping Rule is a model that contains an `MQTT topic pattern`, a `Kafka topic template`, and optionally
+a `Kafka record key`.
+All the incoming MQTT message's topic should match an `MQTT topic pattern` in the ToMaR so that the bridge knows in
```
ppatierno, line 24:
```suggestion
All the incoming MQTT message's topic should match an `MQTT topic pattern` in the ToMaR so that the bridge knows in which Kafka topic to produce this message.
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+which Kafka topic to produce this message.
 This Kafka topic is defined by a template, `Kafka Topic template`.
-However, if the incoming MQTT message's topic does not match any pattern in the ToMaR, the Bridge has a default Kafka topic where the incoming message will be mapped to, known as `messages_default`.
+However, if the incoming MQTT message's topic does not match any pattern in the ToMaR, the Bridge has a default Kafka
+topic where the incoming message will be mapped to, known as `messages_default`.
```
ppatierno, line 28:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
-All the incoming MQTT message's topic should match an `MQTT topic pattern` in the ToMaR so that the bridge knows in which Kafka topic to produce this message.
+The ToMaR is a set of patterns the user provides defining how the MQTT Bridge maps MQTT topic names to Kafka topic
+names.
+A Mapping Rule is a model that contains an `MQTT topic pattern`, a `Kafka topic template`, and optionally
+a `Kafka record key`.
+All the incoming MQTT message's topic should match an `MQTT topic pattern` in the ToMaR so that the bridge knows in
+which Kafka topic to produce this message.
 This Kafka topic is defined by a template, `Kafka Topic template`.
-However, if the incoming MQTT message's topic does not match any pattern in the ToMaR, the Bridge has a default Kafka topic where the incoming message will be mapped to, known as `messages_default`.
+However, if the incoming MQTT message's topic does not match any pattern in the ToMaR, the Bridge has a default Kafka
```
ppatierno, line 27:
```suggestion
However, if the incoming MQTT message's topic does not match any pattern in the ToMaR, the Bridge has a default Kafka topic where the incoming message will be mapped to, known as `messages_default`.
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+     "kafkaKey": "device_$2"
   }
 ]
 ```
 
-The wildcard `#` represents one or more levels in the MQTT topic hierarchy.
-The wildcard `+` represents a single level in the MQTT topic hierarchy.
-It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
+The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic
+hierarchy.
```
ppatierno, line 60:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+     "kafkaTopic": "device_$1_data",
+     "kafkaKey": "device_$2"
   }
 ]
 ```
 
-The wildcard `#` represents one or more levels in the MQTT topic hierarchy.
-The wildcard `+` represents a single level in the MQTT topic hierarchy.
-It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
+The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic
```
ppatierno, line 59:
```suggestion
The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic hierarchy.
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+hierarchy.
+The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT
+topic hierarchy.
```
ppatierno, line 62:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
   }
 ]
 ```
 
-The wildcard `#` represents one or more levels in the MQTT topic hierarchy.
-The wildcard `+` represents a single level in the MQTT topic hierarchy.
-It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
+The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic
+hierarchy.
+The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT
```
ppatierno, line 61:
```suggestion
The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT topic hierarchy.
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+topic hierarchy.
+It worth's mentioning that it is the user's responsibility to adhere to
+the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106)
```
ppatierno, line 64:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+when defining the MQTT topic patterns.
```
ppatierno, line 65:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
 ```
 
-The wildcard `#` represents one or more levels in the MQTT topic hierarchy.
-The wildcard `+` represents a single level in the MQTT topic hierarchy.
-It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
+The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic
+hierarchy.
+The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT
+topic hierarchy.
+It worth's mentioning that it is the user's responsibility to adhere to
```
ppatierno, line 63:
```suggestion
It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106)
+when defining the MQTT topic patterns.
+
+Placeholders in the `KafkaTopic` and `kafkaKey` template are defined using the `$` character followed by a number. The
+number represents the index of the capturing group in the `mqttTopic` pattern.
```
ppatierno, line 68:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
-It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
+The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic
+hierarchy.
+The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT
+topic hierarchy.
+It worth's mentioning that it is the user's responsibility to adhere to
+the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106)
+when defining the MQTT topic patterns.
+
+Placeholders in the `KafkaTopic` and `kafkaKey` template are defined using the `$` character followed by a number. The
```
ppatierno, line 67:
```suggestion
Placeholders in the `kafkaTopic` and `kafkaKey` templates are defined using the `$` character followed by a number.
The number represents the index of the capturing group in the `mqttTopic` pattern.
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+number represents the index of the capturing group in the `mqttTopic` pattern.
+The MQTT Bridge uses the capturing groups in the `mqttTopic` pattern to positionally extract the values that will be
+used to replace the placeholders in the `KafkaTopic` and `kafkaKey` templates.
```
ppatierno, line 70:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+hierarchy.
+The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT
+topic hierarchy.
+It worth's mentioning that it is the user's responsibility to adhere to
+the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106)
+when defining the MQTT topic patterns.
+
+Placeholders in the `KafkaTopic` and `kafkaKey` template are defined using the `$` character followed by a number. The
+number represents the index of the capturing group in the `mqttTopic` pattern.
+The MQTT Bridge uses the capturing groups in the `mqttTopic` pattern to positionally extract the values that will be
```
ppatierno, line 69:
```suggestion
The MQTT Bridge uses the capturing groups in the `mqttTopic` pattern to positionally extract the values that will be used to replace the placeholders in the `kafkaTopic` and `kafkaKey` templates.
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+used to replace the placeholders in the `KafkaTopic` and `kafkaKey` templates.
 
-The parts with curly braces in the MQTT topic pattern are called `placeholders`.
-The MQTT Bridge will extract the values of these placeholders from the MQTT topic and use them to construct the Kafka topic name using the `kafkaTopic` template.
+Let's go through each rule in the above example to understand how the MQTT Bridge uses these rules to map MQTT topics to
+Kafka topics:
```
ppatierno, line 73:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+when defining the MQTT topic patterns.
+
+Placeholders in the `KafkaTopic` and `kafkaKey` template are defined using the `$` character followed by a number. The
+number represents the index of the capturing group in the `mqttTopic` pattern.
+The MQTT Bridge uses the capturing groups in the `mqttTopic` pattern to positionally extract the values that will be
+used to replace the placeholders in the `KafkaTopic` and `kafkaKey` templates.
 
-The parts with curly braces in the MQTT topic pattern are called `placeholders`.
-The MQTT Bridge will extract the values of these placeholders from the MQTT topic and use them to construct the Kafka topic name using the `kafkaTopic` template.
+Let's go through each rule in the above example to understand how the MQTT Bridge uses these rules to map MQTT topics to
```
ppatierno, line 72:
```suggestion
Let's go through each rule in the above example to understand how the MQTT Bridge uses these rules to map MQTT topics to Kafka topics:
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+Kafka topics:
 
-Let's go through each rule in the above example to understand how the MQTT Bridge uses these rules to map MQTT topics to Kafka topics:
+1. MQTT Topic: `building/(\\w+)/room/(\\d{1,4}).*` -> Kafka Topic: `building_$1` with Kafka Key: `room_$2`
 
-1. MQTT Topic: `building/{building}/room/{room}/#` -> Kafka Topic: `building_{building}_room_{room}`
+   This rule maps MQTT topics of the form `building/{some word}/room/{some number with 4 digits}/#` to the Kafka
+   topic `building_$1`.
```
ppatierno, line 78:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
-The parts with curly braces in the MQTT topic pattern are called `placeholders`.
-The MQTT Bridge will extract the values of these placeholders from the MQTT topic and use them to construct the Kafka topic name using the `kafkaTopic` template.
+Let's go through each rule in the above example to understand how the MQTT Bridge uses these rules to map MQTT topics to
+Kafka topics:
 
-Let's go through each rule in the above example to understand how the MQTT Bridge uses these rules to map MQTT topics to Kafka topics:
+1. MQTT Topic: `building/(\\w+)/room/(\\d{1,4}).*` -> Kafka Topic: `building_$1` with Kafka Key: `room_$2`
 
-1. MQTT Topic: `building/{building}/room/{room}/#` -> Kafka Topic: `building_{building}_room_{room}`
+   This rule maps MQTT topics of the form `building/{some word}/room/{some number with 4 digits}/#` to the Kafka
```
ppatierno, line 77:
```suggestion
   This rule maps MQTT topics of the form `building/{some word}/room/{some number with 4 digits}/#` to the Kafka topic `building_$1`.
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+   topic `building_$1`.
+   For example, if the MQTT topic is `building/A/room/1003/floor/2` it will be mapped to the Kafka
+   topic `building_A` with the key `room_1003`.
```
ppatierno, line 80:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+   For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data`.
+   Because the `KafkaKey` is not defined, the key of the Kafka record will be `null`.
 
-    This rule maps MQTT topics of the form `sensors/+/data` to the Kafka topic `sensor_data`. 
-    For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data`.
+3. MQTT Topic: `sensors.*` -> Kafka Topic: `sensor_others` with Kafka Key: `null`
 
-3. MQTT Topic: `sensors/#` -> Kafka Topic: `sensor_others`
+   This rule maps any MQTT topic starting with `sensors/` followed by any number of levels in the hierarchy to the Kafka
+   topic `sensor_others`.
```
ppatierno, line 91:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+   For example, if the MQTT topic is `sensors/temperature/living-room`, it will be mapped to the Kafka
+   topic `sensor_others` with the key `sensor`.
+   Because the `KafkaKey` is not defined, the key of the Kafka record will be `null`.
 
-   This rule maps any MQTT topic starting with `sensors/` followed by any number of levels in the hierarchy to the Kafka topic `sensor_others`.
-   For example, if the MQTT topic is `sensors/temperature/living-room`,  it will be mapped to the Kafka topic `sensor_others`.
+4. MQTT Topic: `devices/([^/]+)/data/(\b(all|new|old)\b)` -> Kafka Topic: `device_$1_data` with Kafka Key: `device_$2`
 
+   This rule maps MQTT topics of the form `devices/{some word}/data/{either all, new or old}` to the Kafka
```
ppatierno, line 98:
```suggestion
   This rule maps MQTT topics of the form `devices/{some word}/data/{either all, new or old}` to the Kafka topic `device_$1_data`.
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+   topic `device_$1_data`.
+   For example, if the MQTT topic is `devices/thermostat/data/all`, it will be mapped to the Kafka
+   topic `device_thermostat_data` with the key `device_all`.
```
ppatierno, line 101:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+   topic `sensor_others` with the key `sensor`.
+   Because the `KafkaKey` is not defined, the key of the Kafka record will be `null`.
 
-   This rule maps any MQTT topic starting with `sensors/` followed by any number of levels in the hierarchy to the Kafka topic `sensor_others`.
-   For example, if the MQTT topic is `sensors/temperature/living-room`,  it will be mapped to the Kafka topic `sensor_others`.
+4. MQTT Topic: `devices/([^/]+)/data/(\b(all|new|old)\b)` -> Kafka Topic: `device_$1_data` with Kafka Key: `device_$2`
 
+   This rule maps MQTT topics of the form `devices/{some word}/data/{either all, new or old}` to the Kafka
+   topic `device_$1_data`.
+   For example, if the MQTT topic is `devices/thermostat/data/all`, it will be mapped to the Kafka
```
ppatierno, line 100:
```suggestion
   For example, if the MQTT topic is `devices/thermostat/data/all`, it will be mapped to the Kafka topic `device_thermostat_data` with the key `device_all`.
```

ppatierno (changes requested):
A first bunch of suggestions on the documentation. It seems you don't really like one sentence per line approach :-) Or you are using a tool screwing everything up .... so I would change it ;-) Another bunch of suggestions are coming, it seems there is a limit on GitHub (at 35).

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+   topic `device_thermostat_data` with the key `device_all`.
+   This example also shows the advantage of using regex in the `mqttTopic` pattern. The last capturing group in
+   the `mqttTopic` should be a word boundary `\b` followed by either `all`, `new` or `old` and anything else will not
```
ppatierno, line 103:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+   match the pattern.
```
ppatierno, line 104:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
 
-   This rule maps any MQTT topic starting with `sensors/` followed by any number of levels in the hierarchy to the Kafka topic `sensor_others`.
-   For example, if the MQTT topic is `sensors/temperature/living-room`,  it will be mapped to the Kafka topic `sensor_others`.
+4. MQTT Topic: `devices/([^/]+)/data/(\b(all|new|old)\b)` -> Kafka Topic: `device_$1_data` with Kafka Key: `device_$2`
 
+   This rule maps MQTT topics of the form `devices/{some word}/data/{either all, new or old}` to the Kafka
+   topic `device_$1_data`.
+   For example, if the MQTT topic is `devices/thermostat/data/all`, it will be mapped to the Kafka
+   topic `device_thermostat_data` with the key `device_all`.
+   This example also shows the advantage of using regex in the `mqttTopic` pattern. The last capturing group in
```
ppatierno, line 102:
```suggestion
   This example also shows the advantage of using regex in the `mqttTopic` pattern.
   The last capturing group in the `mqttTopic` should be a word boundary `\b` followed by either `all`, `new` or `old` and anything else will not match the pattern.
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+   the `mqttTopic` should be a word boundary `\b` followed by either `all`, `new` or `old` and anything else will not
+   match the pattern.
 
 The order in which the rules are defined is important.
 The MQTT Bridge will use the first rule that matches the MQTT topic.
-For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data` because `sensors/+/data` matches the MQTT topic before `sensors/#`. 
-If we swap the positions of the rules, the MQTT Bridge would use the `sensors/#` to map the MQTT topic to the Kafka topic `sensor_others`.
+For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data`
+because `sensors/([^/]+)/data` matches the MQTT topic before `sensors/#`.
```
ppatierno, line 109:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+   topic `device_thermostat_data` with the key `device_all`.
+   This example also shows the advantage of using regex in the `mqttTopic` pattern. The last capturing group in
+   the `mqttTopic` should be a word boundary `\b` followed by either `all`, `new` or `old` and anything else will not
+   match the pattern.
 
 The order in which the rules are defined is important.
 The MQTT Bridge will use the first rule that matches the MQTT topic.
-For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data` because `sensors/+/data` matches the MQTT topic before `sensors/#`. 
-If we swap the positions of the rules, the MQTT Bridge would use the `sensors/#` to map the MQTT topic to the Kafka topic `sensor_others`.
+For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data`
```
ppatierno, line 108:
```suggestion
For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data` because `sensors/([^/]+)/data` matches the MQTT topic before `sensors/#`.
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+because `sensors/([^/]+)/data` matches the MQTT topic before `sensors/#`.
+If we swap the positions of the rules, the MQTT Bridge would use the `sensors.*` to map the MQTT topic to the Kafka
+topic `sensor_others`.
```
ppatierno, line 111:
```suggestion
```

README.md
```diff
@@ -17,62 +17,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+   the `mqttTopic` should be a word boundary `\b` followed by either `all`, `new` or `old` and anything else will not
+   match the pattern.
 
 The order in which the rules are defined is important.
 The MQTT Bridge will use the first rule that matches the MQTT topic.
-For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data` because `sensors/+/data` matches the MQTT topic before `sensors/#`. 
-If we swap the positions of the rules, the MQTT Bridge would use the `sensors/#` to map the MQTT topic to the Kafka topic `sensor_others`.
+For example, if the MQTT topic is `sensors/temperature/data`, it will be mapped to the Kafka topic `sensor_data`
+because `sensors/([^/]+)/data` matches the MQTT topic before `sensors/#`.
+If we swap the positions of the rules, the MQTT Bridge would use the `sensors.*` to map the MQTT topic to the Kafka
```
ppatierno, line 110:
```suggestion
If we swap the positions of the rules, the MQTT Bridge would use the `sensors.*` to map the MQTT topic to the Kafka topic `sensor_others`.
```

ppatierno (changes requested):
Another bunch of doc suggested changes ;-)

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
```diff
@@ -49,13 +52,22 @@ public MqttServerHandler(KafkaBridgeProducer kafkaBridgeProducer) {
...
-            this.mqttKafkaMapper = new MqttKafkaMapper(rules);
+            this.mqttKafkaMapper = new MqttKafkaRegexMapper(rules);
         } catch (IOException e) {
             logger.error("Error reading mapping file: ", e);
         }
         this.kafkaBridgeProducer = kafkaBridgeProducer;
     }
 
+    /**
+     * Transform a MqttPublishMessage's payload into byte array
```
ppatierno, line 63:
```suggestion
     * Transform a MqttPublishMessage's payload into bytes array
```

src/main/java/io/strimzi/kafka/bridge/mqtt/mapper/MappingRule.java
```diff
@@ -57,6 +62,15 @@ public String getMqttTopicPattern() {
         return mqttTopicPattern;
     }
 
+    /**
+     * Get the record key.
```
ppatierno, line 66:
```suggestion
     * Get the record key template.
```

ppatierno (changes requested):
Great work! We are really close. After you addressed my suggestions I will play a little bit more with the code.

src/main/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaMapper.java
```diff
@@ -4,155 +4,38 @@
...
-
-    /**
-     * Helper method for removing the curly brackets from a placeholder.
-     * @param placeholder represents a placeholder in the pattern.
-     * @return a placeholder without the curly brackets.
-     */
-    private String removeBrackets(String placeholder) {
-        return placeholder.replaceAll("\\{+|\\}+", "");
-    }
+    protected abstract void buildOrCompilePatterns();
```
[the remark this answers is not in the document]
antonio-pedro99: I added javadocs before, let me see if the recent commit brought here this time.

ppatierno (changes requested):

ppatierno:
I left some comments, I think we are on the right path and close to merge.
My only concern is about allowing the user to use both `sensors/[^/]+/data` and `sensors/([^/]+)/data` which look to be similar but with the former you cannot use placeholder variables that you can use with the latter.
We are saying that regex "group"(s) can be used for placeholder variables. Should we make it clear in the doc?

src/main/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaMapper.java
```diff
@@ -4,155 +4,38 @@
...
-    // matches any character after the string. Used to replace the # in the mqtt pattern.
-    private static final String MULTIPLE_LEVEL_WILDCARD_REGEX = ".*";
-    private final List<MappingRule> rules;
-    private final List<Pattern> patterns = new ArrayList<>();
-    private final Pattern placeholderPattern;
-
-    /**
-     * Constructor
-     * Creates a new instance of MqttKafkaMapper.
-     */
```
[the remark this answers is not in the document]
antonio-pedro99: Interesting, I did not remember deleting that. putting back.

src/test/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaRegexMapperTest.java
```diff
@@ -0,0 +1,300 @@
...
+
+        MqttKafkaRegexMapper mapper = new MqttKafkaRegexMapper(rules);
+
+        // Test building/(\\d+)/floor/(\\d+).*
+        MappingResult mappingResult = mapper.map("building/14/floor/25");
+
+        assertThat("building/(\\d+)/floor/(\\d+).* should be mapped to building_$1_$2",
+                mappingResult.kafkaTopic(), is("building_14"));
+
+        assertThat("The key for building_$1 should be expanded to building_25",
```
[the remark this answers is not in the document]
antonio-pedro99: can't believe I missed this too :(

README.md
```diff
@@ -18,68 +18,83 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
-    "kafkaTopic": "sensor_others"
+     "mqttTopic": "sensors.*",
+     "kafkaTopic": "sensor_others"
+  },
+  {
+    "mqttTopic": "devices/([^/]+)/data/(\b(all|new|old)\b)",
+     "kafkaTopic": "device_$1_data",
+     "kafkaKey": "device_$2"
   }
 ]
```
kyguy, line 52: Would it be worth keeping a separate example for the MqttKafkaSimpleMapper in addition to the MqttKafkaRegexMapper?

kyguy (commented):
Is the mapper type configurable by the user? Or is it set to always use the MqttKafkaRegexMapper?

README.md
ppatierno: Yeah I mentioned that in the past. Even if not used I would leave the placeholders based example in the test folder.

ppatierno:
@kyguy I would leave as it is today and see if we "can" use the other one in a customizable way. Maybe at some point we could also allow the users to develop their own mapper and plug it into the bridge.

README.md
antonio-pedro99: I have added the example under `test/resources` folder.

README.md
```diff
@@ -18,68 +18,83 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
 
-The wildcard `#` represents one or more levels in the MQTT topic hierarchy.
-The wildcard `+` represents a single level in the MQTT topic hierarchy.
+The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic hierarchy.
+The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT topic hierarchy.
 It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
 
-The parts with curly braces in the MQTT topic pattern are called `placeholders`.
-The MQTT Bridge will extract the values of these placeholders from the MQTT topic and use them to construct the Kafka topic name using the `kafkaTopic` template.
+Placeholders in the `KafkaTopic` and `kafkaKey` templates are defined using the `$` character followed by a number.
```
ppatierno, line 59:
```suggestion
Placeholders in the `kafkaTopic` and `kafkaKey` templates are defined using the `$` character followed by a number. 
```

README.md
```diff
@@ -18,68 +18,83 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT topic hierarchy.
 It worth's mentioning that it is the user's responsibility to adhere to the [MQTT 3.1.1 naming conventions](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106) when defining the MQTT topic patterns.
 
-The parts with curly braces in the MQTT topic pattern are called `placeholders`.
-The MQTT Bridge will extract the values of these placeholders from the MQTT topic and use them to construct the Kafka topic name using the `kafkaTopic` template.
+Placeholders in the `KafkaTopic` and `kafkaKey` templates are defined using the `$` character followed by a number. 
+The number represents the index of the `capturing group` in the `mqttTopic` pattern. `Please note that the index starts from 1 and not 0`.
+The MQTT Bridge uses the `capturing groups` in the `mqttTopic` pattern to positionally extract the values that will be used to replace the placeholders in the `KafkaTopic` and `kafkaKey` templates.
```
ppatierno, line 61:
```suggestion
The MQTT Bridge uses the `capturing groups` in the `mqttTopic` pattern to positionally extract the values that will be used to replace the placeholders in the `kafkaTopic` and `kafkaKey` templates.
```

ppatierno (commented):
Last two nits then I am ready to approve.

README.md
antonio-pedro99: I have added this to my other changes to avoid having 1 PR to each change.

antonio-pedro99:
@ppatierno The code working as expected?

src/test/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaRegexMapperTest.java
```diff
@@ -0,0 +1,300 @@
...
+        MqttKafkaRegexMapper mapper = new MqttKafkaRegexMapper(rules);
+
+        // Test building/(\\d+)/floor/(\\d+).*
+        MappingResult mappingResult = mapper.map("building/14/floor/25");
+
+        assertThat("building/(\\d+)/floor/(\\d+).* should be mapped to building_$1",
+                mappingResult.kafkaTopic(), is("building_14"));
+
+        assertThat("The key for building_$1 should be expanded to room_25",
+                mappingResult.kafkaKey(), is("room_25"));
```
kyguy, line 68:
```suggestion
                mappingResult.kafkaKey(), is("floor_25"));
```

kyguy (approved):
Solid work @antonio-pedro99,  just found a minor test issue but everything else is good

src/test/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaRegexMapperTest.java
ppatierno: Yeah the test is failing locally. The Azure pipeline will help on next PRs ;-)

ppatierno (changes requested):

src/test/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaRegexMapperTest.java
```diff
@@ -0,0 +1,300 @@
...
+
+        MqttKafkaRegexMapper mapper = new MqttKafkaRegexMapper(rules);
+
+        // Test building/(\\d+)/floor/(\\d+).*
+        MappingResult mappingResult = mapper.map("building/14/floor/25");
+
+        assertThat("building/(\\d+)/floor/(\\d+).* should be mapped to building_$1",
+                mappingResult.kafkaTopic(), is("building_14"));
+
+        assertThat("The key for building_$1 should be expanded to room_25",
```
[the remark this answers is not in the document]
ppatierno: @antonio-pedro99 Mr. Kyle fixed the check line on the `is` predicate, not on the assert message.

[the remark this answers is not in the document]
antonio-pedro99: Oh, I have got it fixed locally. It is coming with the next commit just .

README.md
```diff
@@ -52,8 +52,11 @@ The following is an example of a valid ToMaR:
 ]
 ```
 
-The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic hierarchy.
-The expression `([^/]+)` is used to represent the wildcard `+`, which in turn represents a single level in the MQTT topic hierarchy.
+- The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic hierarchy.
+However, there are some cases that can lead to unexpected behavior when using the `.*` wildcard. Therefore, you should note the following:
```
ppatierno, line 56:
```suggestion
However, there are some cases that can lead to unexpected behavior when using the `.*` wildcard.
Therefore, you should note the following:
```

README.md
```diff
@@ -52,8 +52,11 @@ The following is an example of a valid ToMaR:
...
+  - You can not use the `.*` wildcard in capturing groups. For example, the pattern `building/(\\w+)/room/(\\d{1,4})/(.*)` is invalid because it will capture the whole subtopic of the pattern, and `$3` placeholder might include `/` characters, which will lead to an invalid Kafka topic name.
```
ppatierno, line 57:
```suggestion
  - You cannot use the `.*` wildcard in capturing groups. 
  For example, the pattern `building/(\\w+)/room/(\\d{1,4})/(.*)` is invalid because it will capture the whole subtopic of the pattern, and `$3` placeholder might include `/` characters, which will lead to an invalid Kafka topic name.
```

README.md
```diff
@@ -52,8 +52,11 @@ The following is an example of a valid ToMaR:
...
+  - You can use the `.*` wildcard without a preceding `/` character, but it is not equivalent to when it is preceded by `/`. For example, the pattern `building.*` will match `building`, `building/room`, `buildingroom`, and so on. On the other hand, the pattern `building/.*` will match `building`, `building/room`, `building/floor/room`, and so on, but not `buildingroom`.
```
ppatierno, line 58:
```suggestion
  - You can use the `.*` wildcard without a preceding `/` character, but it is not equivalent to when it is preceded by `/`. 
  For example, the pattern `building.*` will match `building`, `building/room`, `buildingroom`, and so on. 
  On the other hand, the pattern `building/.*` will match `building`, `building/room`, `building/floor/room`, and so on, but not `buildingroom`. 
```

src/main/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaRegexMapper.java
```diff
@@ -54,7 +57,22 @@ public MappingResult map(String mqttTopic) {
...
+                        MqttKafkaRegexMapper.WILDCARD_REGEX;
+                this.patterns.add(Pattern.compile(regex));
+            } else if(rule.getMqttTopicPattern().endsWith("("+PATTERN_MULTI_LEVEL_WILDCARD_REGEX+")")){
+                throw new IllegalArgumentException("The pattern " + rule.getMqttTopicPattern() + " is not valid. You should not use .* in capture groups.\n" +
+                        "Example of correct use of .* are: building.*, building/.*, and etc.\n"+
+                        "Please refer to the documentation for more information.");
+            } else {
+                this.patterns.add(Pattern.compile(rule.getMqttTopicPattern()));
+            }
+        }
```
ppatierno, line 75: I know you went further after our offline discussion but now I see this code adding more complexity for no additional value imho.
I think that we should revert back, by leaving the user the freedom to set a regex which we can just compile here (as the code was before).
Documenting that `buildings.*` and `buildings/.*` are two different things is important anyway.
Wdyt @antonio-pedro99 @kyguy ?

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaRegexMapper.java
antonio-pedro99: maybe I should make the code cleaner?
I am apologist that the users should use `buildings/.*` to mean `buildings/#`
The previous implementation would not make this behaves exactly how it should behave when the topic is `buildings/#`

On the other hand, `building.*` looks more like `building#` (which is not a valid topic) and it would be confusing to the users.

And last but not least, if you noticed was replacing `.*` with `(?:\\/.*)?$` this the correct regex to deal with, because even when we use something like `building(?:/.*)?$`, `buildingslkmnskmskms` will never match the expression. However, this expression seems to big :( 

If you are fine, maybe we can go ahead and use directly `(?:/.*)?$` instead of `.*`

kyguy: I don't think defending against misuse of regex is worth the time or complexity for this PR. Maybe it is something we could revisit after mapping out all possible cases of misuse.

For now, I think we should compile the regex provided from the user directly as Paolo suggested, leaving the regex as a responsibility to the user. That was the original purpose of the regex mapper anyway right? To give users flexibility and save the application from the complexity.

Maybe we could provide some examples of some common (proper) regex patterns to use and list some common regex patterns to avoid.

README.md
```diff
@@ -48,18 +48,26 @@ The following is an example of a valid ToMaR:
...
 ]
 ```
 
-- The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic hierarchy.
-However, there are some cases that can lead to unexpected behavior when using the `.*` wildcard.
+The expressions `.*` and `(?:\\/.*)?$` are used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic hierarchy.
+However, there are some cases that can lead to unexpected behavior when using the these wildcards interchangeably . 
 Therefore, you should note the following:
-  - You cannot use the `.*` wildcard in capturing groups. 
+  - You can not use the `.*` wildcard in capturing groups.
```
ppatierno, line 63:
```suggestion
  - You cannot use the `.*` wildcard in capturing groups. 
```

README.md
```diff
@@ -48,18 +48,26 @@ The following is an example of a valid ToMaR:
...
+      "mqttTopic": "locations/([^/]+)(?:\\/.*)?$",
+      "kafkaTopic": "locations",
+      "kafkaKey": "location_$1"
   }
 ]
 ```
 
-- The expression `.*` is used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic hierarchy.
-However, there are some cases that can lead to unexpected behavior when using the `.*` wildcard.
+The expressions `.*` and `(?:\\/.*)?$` are used to represent the wildcard `#`, which in turn represents one or more levels in the MQTT topic hierarchy.
```
ppatierno, line 60: I think a user can understand `.*` but can you provide more details about the `(?:\\/.*)?$` to explain why and how it's different from `.*`. Why the user should use one or the other?

antonio-pedro99: I explained down below.

ppatierno: Tbh you are comparing the two with different examples, it doesn't help the user. See my comment below.

README.md
```diff
@@ -18,68 +18,98 @@ As a part of the bridge, a Kafka producer will be responsible for producing mess
...
+However, there are some cases that can lead to unexpected behavior when using the these wildcards interchangeably . 
+Therefore, you should note the following:
+  - You cannot use the `.*` wildcard in capturing groups. 
+  For example, the pattern `building/(\\w+)/room/(\\d{1,4})/(.*)` is invalid because it will capture the whole subtopic of the pattern, and `$3` placeholder might include `/` characters, which will lead to an invalid Kafka topic name.
+  - You can use the `.*` wildcard without a preceding `/` character, but it is not equivalent to when it is preceded by `/`. 
+  For example, the pattern `building.*` will match `building`, `building/room`, `buildingroom`, and so on. 
+  On the other hand, the pattern `building/.*` will match `building/room`, `building/floor/room`, and so on, but not `buildingroom` and `building/`. 
+  - You can also use the `(?:\\/.*)?$` wildcard to match the whole subtopic level of the pattern. 
+  For example, the pattern `locations/([^/]+)(?:\\/.*)?$` will match `locations/city/luanda/angola`, `locations/city`, `locations/city/`, and so on.
```
ppatierno, line 69: If you want to make a good comparison between `.*` and `(?:\\/.*)?$` you should use the same example and show how each of them is able to match differently.

ppatierno (approved):
Finally, LGTM. Thanks! :-)

PR #29, opened 2023-07-19
style: add unix linefeed

antonio-pedro99:
Not sure if you have any specific preference for `line endings`. 
Please ignore if it is not important.

GitHub warning:
image


ppatierno:
Yeah we don't have strict rules, we can live with what we have now.

PR #36, opened 2023-07-27
Highlighting MQTT to Kafka only in the landing page

### LINKED ISSUE #35: Question about scope
Hey all,

I'm wondering if the scope is only to map mqtt->kafka or also from kafka->mqtt? I don't see any mention of kafka->mqtt, but in any case I think it would be good to mention that explicitly.

ppatierno:
This PR fixes #35 making clear that the current MQTT to Kafka bridge just enables the one-way communication from MQTT to Kafka (and not viceversa).


antonio-pedro99 (approved):
LGTM

kyguy (approved):

PR #38, opened 2023-07-31
tests: Added integration test

antonio-pedro99:
This PR adds integration tests for the Bridge.

With the help of [Strimzi TestContainer](https://github.com/strimzi/test-container), we can bring a Kafka Cluster ready for use. 
Before all the tests, this cluster is initialized and a Kafka Consumer client subscribes to a topic right after creating its instance. 
The MQTT starts once the cluster is ready, this Bridge now runs in a different thread to make sure it does not block the main thread.

Once the Bridge is up, we first test if a paho based MQTT clients can connect to the Bridge.
I also added tests for message publishing and consuming from the Kafka consumer.


pom.xml
```diff
@@ -78,18 +80,36 @@
             <artifactId>kafka-clients</artifactId>
             <version>${kafka.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>org.eclipse.paho</groupId>
+            <artifactId>org.eclipse.paho.client.mqttv3</artifactId>
+            <version>${eclipse-paho.version}</version>
+        </dependency>
```
ppatierno, line 88: You should move this to be a test dependency only.

src/main/java/io/strimzi/kafka/bridge/mqtt/Main.java
```diff
@@ -51,8 +51,8 @@ public static void main(String[] args) {
 
             // start the MQTT server
             mqttServer.start();
-        } catch (IOException | ParseException e) {
-            logger.error("Error starting the MQTT server: ", e);
+        } catch (Exception e) {
+                logger.error("Error starting the MQTT server: ", e);
```
ppatierno, line 55: Why changing indentation?

src/test/java/io/strimzi/kafka/bridge/mqtt/clients/KafkaConsumerClient.java
```diff
@@ -0,0 +1,68 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.time.Duration;
+import java.util.Collections;
+import java.util.Properties;
+
+/**
+ * A Kafka consumer client
+ */
+public class KafkaConsumerClient {
```
ppatierno, line 21: Why we need this class it's not doing that much, just wrapping and calling the underneath Kafka consumer client. Any specific reasons for it?

src/test/java/io/strimzi/kafka/bridge/mqtt/mqtt/MqttBridgetIT.java
```diff
@@ -0,0 +1,179 @@
...
+        });
+        prepareServerThread.start();
+        Thread.sleep(5000);
+    }
+
+    /**
+     * Create a new client id for each test randomly after each test
+     */
+    @AfterEach
+    public void afterEach() {
```
ppatierno, line 107: Should be this a beforeEach? I would expect that before starting each test a new clientid is generated for such a test, or? Generating after is counterintuitive imho.

src/test/java/io/strimzi/kafka/bridge/mqtt/clients/KafkaConsumerClient.java
antonio-pedro99: Because of the consumer properties? I want to leave the MqttBridgeIT cleaner. So we wrap everything needed for the consumer in a class.
To be honest I wanted to do the same for the Mqtt client, its configuration is simpler.

src/test/java/io/strimzi/kafka/bridge/mqtt/mqtt/MqttBridgetIT.java
antonio-pedro99: Yes, it works! 
I was using afterEach because the clientId was already initialized.

What I am saying is somthing like
```java
 private String mqttClientId = "mqttClientId-" + new Random().nextInt();

@AfterEach
clientId += new Random().nextInt(10);
````
I think they are the same, aren't they?

Comment (antonio-pedro99) [reply src/test/java/io/strimzi/kafka/bridge/mqtt/mqtt/MqttBridgetIT.java:107]: But the @BeforeEach seems to be simpler. I will go with that.

File: src/main/java/io/strimzi/kafka/bridge/mqtt/Main.java (hunk 2 above)
Comment (antonio-pedro99) [reply src/main/java/io/strimzi/kafka/bridge/mqtt/Main.java:55]: No specif reason, I think I pressed tab mistakenly

File: pom.xml [hunk 5]
```diff
@@ -78,18 +80,36 @@
...
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>${junit.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <version>5.8.2</version>
```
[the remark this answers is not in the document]
antonio-pedro99: I was thinking maybe I should migrate all tests from junit 4 to 5 in a different PR, to make your review easier.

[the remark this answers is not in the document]
antonio-pedro99: I have opened a PR updating the dependencies. After merging #39, I will rebase the current branch.

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
```diff
@@ -0,0 +1,252 @@
...
+
+
+    /**
+     * Start the kafka Cluster
+     * Start the MQTT bridge before all tests
+     */
+    @BeforeAll
+    public static void beforeAll() throws InterruptedException {
+        final Map<String, Object> config = new HashMap<>();
+        String kafkaBootstrapServers = "localhost:9092";
```
ppatierno, line 78: I am not sure this default initialization is really needed. If the Kafka container doesn't start (and we just log an exception) you are expecting to connect a "localhost:9092" which actually doesn't exist, while the test has to fail earlier I would say.

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
```diff
@@ -0,0 +1,252 @@
...
+                mqttBridge.start();
+            } catch (InterruptedException interruptedException) {
+                // in this case, we are interrupting the thread, so we don't need to do anything
+                logger.info("Shutting down the server");
+            } catch (Exception e) {
+                logger.error("Exception occurred", e);
+            }
+        });
+        prepareServerThread.start();
+        Thread.sleep(5000);
```
ppatierno, line 129: What's the reason for this sleep? We don't like sleep :-)
If you have a sync problem waiting for the MQTT server being started we have to deal with it better.
Maybe it would mean having the bridge `start` returning a `CompletableFuture` and wait for it?

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    /**
+     * Randomly create a new client id before each test
+     */
+    @BeforeEach
+    public void createClientId() {
```
ppatierno, line 136:
```suggestion
    public void generateMqttClientId() {
```

just to be clear it's about MQTT.
Also, do we need to story the MQTT client id in a local member of the class and having this as @BeforeEach? Can't make this private and call it in each test, or when needed?

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
```diff
@@ -0,0 +1,252 @@
...
+                ),
+                new StringDeserializer(),
+                new StringDeserializer()
+        );
+
+        // consumer client subscribes to the kafka topic
+        kafkaConsumerClient.subscribe(Collections.singletonList(KAFKA_TOPIC));
+
+        // start the bridge server in a separate thread
+        prepareServerThread = new Thread(() -> {
```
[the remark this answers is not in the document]
antonio-pedro99: Before this, I was using a CountDownLatch to coordinate the thread. That's why I called this thread prepareServerThread, because it would have the server execution.
But it was not working as expected, so removed the CountDownLatch and forgot to rename it according.

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
antonio-pedro99: good catch! I will just throw the exception.

kyguy (approved):

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
antonio-pedro99: I tried with CountDownLatch before, as I said in the previous comment. Maybe I was not using it appropriately. Let me try it here.
Yeah, we do not want to sleep.

If this sleep is not good, then sleep in @AfterAll should be removed too.

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
```diff
@@ -0,0 +1,252 @@
...
+            // Set up options for the connection
+            MqttConnectOptions options = new MqttConnectOptions();
+
+            IMqttToken conn = client.connectWithResult(options);
+
+            assertThat("The session present flag should be false",
+                    conn.getSessionPresent(), is(false));
+
+            assertThat("The connection return code should be 0",
+                    conn.getMessageId(), is(0));
```
ppatierno, line 155: I am not sure that the id of the message back is meaningful. It should be 0 for sure because it's the first message.
I would check that it's receiving a CONNACK instead.
So using something like this:

```java
assertThat("The connection return message should be CONNACK",
                    conn.getResponse().getType(), is(MESSAGE_TYPE_CONNACK));
```

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
```diff
@@ -0,0 +1,252 @@
...
+
+        MqttConnectOptions options = new MqttConnectOptions();
+        // Connect all clients and test the connection
+        clients.stream().unordered().parallel().forEach(client -> {
+            try {
+                IMqttToken conn = client.connectWithResult(options);
+                assertThat("The session present flag should be false",
+                        conn.getSessionPresent(), is(false));
+                assertThat("The connection return code should be 0",
+                        conn.getMessageId(), is(0));
```
ppatierno, line 189: Ditto as above about checking CONNACK instead.

ppatierno (changes requested):

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
antonio-pedro99: I was actually trying to do the same. I judged that this type is something else, and the message id is the CONNACK.
Thanks for pointing out.

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
```diff
@@ -0,0 +1,252 @@
...
+                    conn.getSessionPresent(), is(false));
+
+            assertThat("The connection return code should be 0",
+                    conn.getMessageId(), is(0));
+
+            assertThat("The client should be connected",
+                    client.isConnected(), is(true));
+
+            assertThat("The client id should be " + mqttClientId,
+                    client.getClientId(), is(mqttClientId));
```
ppatierno, line 161: this check doesn't make much sense. It's not something you are getting back in the connection but you set it a few lines above on client creation. If you follow the constructor, it's just setting an internal member field. So you are testing that the constructor worked :-D Just remove it ;-)

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
ppatierno: In MQTT each message has an id just as a sequential number. The type defines which message it is (CONNACK, PUBLISH, ...).

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
antonio-pedro99: Agree

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
antonio-pedro99: @ppatierno I do not think using Latch fixes the issue.
Maybe I am missing something.

```java
....inside the thread block
// Do some task here
mqttServer.start();

// Signal that the task is completed
latch.countDown();
```
```
The latch will never count down because `.start()` blocks the thread. Of course, there is no sense to count down the latch before starting the server.

Comment (ppatierno) [reply src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java:129]: As I read the `start` method ...

```java
ChannelFuture channelFuture = this.serverBootstrap.bind(host, port).sync();
channelFuture.channel().closeFuture().sync();
```

the first `sync` on the binding is actually waiting synchronously for the binding to be done. So you are actually making an async operation by Netty (the binding) to be sync.
Then next sync actually block the thread where you start the bridge (this is why in the tests you needed a new thread).
It means that in your tests the `start` is a blocking call and you will never be able to use a `CountDownLatch.countDown` right after the start, because it will never executed until the bridge is closed.
I think this is the reason why it didn't work.

So at this point I think the `start` method should be slightly changed. You should waiting for the bind and then completing a future returned by the `start` method itself to unlock the caller.
Right after that you can sync on the closeFuture and actually blocking the bridge thread.

antonio-pedro99: thanks sensei

pom.xml
```diff
@@ -98,5 +100,16 @@
...
+        <dependency>
+            <groupId>io.strimzi</groupId>
+            <artifactId>strimzi-test-container</artifactId>
+            <version>${test-container.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.paho</groupId>
+            <artifactId>org.eclipse.paho.client.mqttv3</artifactId>
+            <version>${eclipse-paho.version}</version>
```
ppatierno, line 112:
```suggestion
            <version>${eclipse-paho.version}</version>
            <scope>test</scope>
```

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
```diff
@@ -0,0 +1,238 @@
...
+        CountDownLatch latch = new CountDownLatch(1);
+
+        EventLoopGroup bossGroup = new NioEventLoopGroup();
+        EventLoopGroup workerGroup = new NioEventLoopGroup();
+
+        mqttBridge = new MqttServer(bridgeConfig, bossGroup, workerGroup, ChannelOption.SO_KEEPALIVE);
+        mqttBridge.start();
+
+        latch.countDown();
+        latch.await();
```
ppatierno, line 123: Actually, I made the `start` method to be sync ... so no need for the `CountDownLatch` at all :-D My bad!

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    /**
+     * Randomly generate a new client id before each test
+     */
+    @BeforeEach
+    public void generateClientId() {
+        mqttClientId = "mqtt-client-" + new Random().nextInt(20);
+    }
```
ppatierno, line 132: still thinking you don't need this to be a `@BeforeEach`. You can make this private and then calling this method each time you need an MQTT client id in each test. So you don't need the private field to store it.

ppatierno (changes requested):

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
antonio-pedro99: Is there any drawback? 

For me it's good using @BeforeEach to avoid the need of calling generateClientId(); all the time.

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
antonio-pedro99: I see

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
ppatierno: So something like this ...

```java
private String generateClientId() {
     return "mqtt-client-" + new Random().nextInt(20);
}
```

remove the `mqttClientId` from the class.

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * Test the client's connection to the Bridge
+     */
+    @Test
+    public void testConnection() throws MqttException {
+        try (MqttClient client = new MqttClient(MQTT_SERVER_URI, mqttClientId)) {
```
ppatierno, line 139: ... and call `generateClientId()` directly here instead of referencing `mqttClientId`.

ppatierno (changes requested):

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
antonio-pedro99: Looks better.

ppatierno: > Is there any drawback? For me it's good using https://github.com/beforeeach to avoid the need of calling generateClientId(); all the time.

It's anyway called all the time. And you are storing the result in a field member which you don't really need which can also drives to side effects :-)

antonio-pedro99: I would call it `generateAndGetClientId()` or simpler `getRandomClientId()`

antonio-pedro99: yes, got you!

ppatierno (approved):

PR #39, opened 2023-08-03
chore: migrate from Junit 4 to Junit 5

antonio-pedro99:
Trivial PR to migrate from Junit 4 to Junit 5.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,11 +15,12 @@
         <netty.version>4.1.94.Final</netty.version>
         <log4j.version>2.17.2</log4j.version>
         <slf4j.version>1.7.21</slf4j.version>
-        <junit.version>4.13.2</junit.version>
+        <junit.version>5.8.2</junit.version>
         <jackson.version>2.15.2</jackson.version>
         <commons-cli.version>1.5.0</commons-cli.version>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -91,5 +92,11 @@
             <version>${mockito.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest</artifactId>
+            <version>${hamcrest.version}</version>
+            <scope>test</scope>
+        </dependency>
```
ppatierno, line 100: What's the reason for this? I don't see any `org.hamcrest` classes used.

antonio-pedro99: We are using for assertions like `assertThat`, and matchers like `is`.
I did not expect that `org.junit.jupter` comes with assertions, maybe should be better to use them instead of using `org.hamrest`?

ppatierno: No sorry, my mistake. You were using it and we use it even in other Strimzi projects so it's fine with me.

ppatierno (approved):

kyguy (approved):

PR #43, opened 2023-08-04
Disable persistence on MQTT Paho clients

ppatierno:
The MQTT Paho client uses file persistence by default for storing information about sessions.
It creates some folders like this `mqtt-client-19-tcplocalhost1883` in the root including a `.lck` file.
Of course, we don't need clients persistence in our tests and having this folders is useless.
This PR disable such a persistence.


kyguy (approved):

antonio-pedro99 (approved):
LGTM

PR #44, opened 2023-08-04
package the application

antonio-pedro99:
Added the code to package the application.


kyguy (commented):
Ran `mvn package && mvn install` to package, then unzipped the package and ran the application without errors (other than missing config-map and mapping-rules file, which is expected)

A couple of questions:
(1) Should we provide a default config-map + mapping rules to use as part of the run script in the same manner that a logging config is provided?
(2) Are these packaging commands going to be added to a centralized Makefile?

antonio-pedro99:
@kyguy 

(1) I think we should not make the config-file a mandatory option. Our implementation also looks for env names for the bridge configuration. 

This could be trivial, but as we have a default Kafka topic, maybe providing the mapping rules should be optional and all topics will be mapped to the default Kafka topic, but this might lead us to some issues, the one we want to solve: `exploding the Kafka topic`. 


(2) AFAIK,  yes, we are adding this In latter PR.

bin/mqtt_bridge_run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -x
+
+# Find my path to use when calling scripts
+MYPATH="$(dirname "$0")"
+
+# Configure logging
+if [ -z "$KAFKA_BRIDGE_LOG4J_OPTS" ]
```
ppatierno, line 8:
```suggestion
if [ -z "$MQTT_BRIDGE_LOG4J_OPTS" ]
```

bin/mqtt_bridge_run.sh
```diff
@@ -0,0 +1,13 @@
...
+then
+      KAFKA_BRIDGE_LOG4J_OPTS="-Dlog4j2.configurationFile=file:${MYPATH}/../config/log4j2.properties"
```
ppatierno, line 10:
```suggestion
      MQTT_BRIDGE_LOG4J_OPTS="-Dlog4j2.configurationFile=file:${MYPATH}/../config/log4j2.properties"
```

bin/mqtt_bridge_run.sh
```diff
@@ -0,0 +1,13 @@
...
+fi
+
+exec java $KAFKA_BRIDGE_LOG4J_OPTS -classpath "${MYPATH}/../libs/*" io.strimzi.kafka.bridge.mqtt.Main "$@"
```
ppatierno, line 13:
```suggestion
exec java $MQTT_BRIDGE_LOG4J_OPTS -classpath "${MYPATH}/../libs/*" io.strimzi.kafka.bridge.mqtt.Main "$@"
```

pom.xml
```diff
@@ -135,6 +137,37 @@
...
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>${maven-jar-plugin.version}</version>
+                <configuration>
+                    <archive>
+                        <manifest>
+                            <mainClass>io.strimzi.kafka.bridge.mqtt.Main</mainClass>
```
ppatierno, line 147:
```suggestion
                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
```

ppatierno: We don't need the `Main` class specified here. It's passed on the command line in the run script.

src/main/assembly/zip.xml
```diff
@@ -0,0 +1,39 @@
...
+        <fileSet>
+            <directory>${project.basedir}/bin</directory>
+            <outputDirectory>/bin</outputDirectory>
+            <fileMode>0755</fileMode>
+        </fileSet>
+        <fileSet>
+            <directory>${project.basedir}</directory>
+            <includes>
+                <include>README*</include>
+                <include>LICENSE*</include>
```
ppatierno, line 26:
```suggestion
                <include>LICENSE*</include>
                <include>CHANGELOG*</include>
```

Including CHANGELOG, it will come with a first release and we could forget about it.

ppatierno (changes requested):

ppatierno:
Regarding the `--config-file` option, it's better to have it mandatory. Even with the Strimzi HTTP bridge experience no one is using env vars and also the Strimzi operator deploy the configuration via the `application.properties` file.

Regarding to provide a default mapping rules, I can't imagine which mapping it should do because it's too much user application related.
An idea I have in mind is providing a mapping rules file with an empty rules? How does the bridge behaves in this case? Do we have a test? It would be great if in such a situation, it would just map to the default topic. It would allow the bridge to work with a default mapping rules file we can provide as part of the config folder.

ppatierno:
@antonio-pedro99 I did just a try ... providing a mapping rules file with just an empty json array, enables the mapping to the default topic. I would do that by adding such a JSON file in the config folder.

antonio-pedro99:
@ppatierno providing an empty json array is also wise. Maybe we can specify it in the docs.

src/main/assembly/assembly.xml
```diff
@@ -0,0 +1,40 @@
+<assembly xmlns="http://maven.apache.org/ASSEMBLY/2.2.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/ASSEMBLY/2.2.0 https://maven.apache.org/xsd/assembly-2.2.0.xsd">
+    <id>zip</id>
```
ppatierno, line 3: you missed changing this id as well in `assembly`

ppatierno (changes requested):

src/main/assembly/assembly.xml
antonio-pedro99: thanks for pointing it out.

ppatierno (approved):
LGTM!

kyguy (approved):

PR #48, opened 2023-08-13
Add the published MQTT topic in the produced Kafka record header

ppatierno:
This trivial PR adds the published MQTT topic to the produced Kafka record header so the Kafka consumer can get where the message was published.


antonio-pedro99 (approved):
LGTM

kyguy (approved):

PR #49, opened 2023-08-13
Refactor MQTT host and port properties

ppatierno:
This PR removes the `server` prefix for MQTT host and port to stick with the same pattern we have with the current Strimzi HTTP bridge.
It's anyway clear what's the purpose of host and port, it could be server only because the bridge doesn't have any client part.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,8 +166,8 @@ A valid configuration file should look like this:
     # Bridge configuration
     bridge.id=my-bridge
     # MQTT configuration
-    mqtt.server.host=localhost
-    mqtt.server.port=1883
+    mqtt.host=localhost
+    mqtt.port=1883
     # Kafka configuration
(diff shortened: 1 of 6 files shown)


antonio-pedro99 (commented):
@ppatierno that makes sense. 

But should we also change the ConfigMap? Because I see it is still using the prefix.

ppatierno:
> But should we also change the ConfigMap? Because I see it is still using the prefix.

@antonio-pedro99 good catch! I also updated some other places (i.e. README)

antonio-pedro99:
> > But should we also change the ConfigMap? Because I see it is still using the prefix.
> 
> @antonio-pedro99 good catch! I also updated some other places (i.e. README)

Yes yes. 
We are ready now.

antonio-pedro99 (approved):

kyguy (approved):

PR #50, opened 2023-08-29
Added make targets for release

ppatierno:
This PR adds the make targets for a release and updates the next version to be 0.1.0 (SNAPSHOT for now).


scholzj (commented):
Should you also:
* Add creation of ZIP / TAR.GZ packages with the installation files?
* Mirror the usual structure we have in our projects with `./packaging` etc.?
* Do you plan to create the Azure pipelines for it in a separate PR? (if needed, https://github.com/strimzi/kafka-access-operator/pull/31 should give you the blueprint for the Azure pipelines I guess)

ppatierno:
> Add creation of ZIP / TAR.GZ packages with the installation files?
Mirror the usual structure we have in our projects with ./packaging etc.?

I have just addressed these two points.

> Do you plan to create the Azure pipelines for it in a separate PR? (if needed, https://github.com/strimzi/kafka-access-operator/pull/31 should give you the blueprint for the Azure pipelines I guess)

As said in the other opened PR, I plan to update Azure pipelines with different PR(s).

Makefile
```diff
@@ -3,9 +3,57 @@ include ./Makefile.maven
 include ./Makefile.docker
 
 PROJECT_NAME ?= mqtt-bridge
+GITHUB_VERSION ?= main
```
scholzj, line 6: Is this used for anything?

scholzj (approved):
I guess the advantage of doing it all at once is that you can actually test that it works. But if you wanna do it separately, then fine with me. I left one more nit. LGTM otherwise I guess.

Makefile
ppatierno: Good catch. I was getting code from the HTTP bridge. I should double check if it's used there somehow or we can remove it on both repos.

kyguy (approved):

Makefile
ppatierno: @scholzj I noticed we are using this in the strimzi operator repo only to set a github version during the documentation generation with asciidoctor. Other repositories have the same env var declared but it's not used at all (I checked drain cleaner, http bridge, access operator). We should remove that env var from those repos. Removing from here for now.

ppatierno:
> I guess the advantage of doing it all at once is that you can actually test that it works. But if you wanna do it separately, then fine with me. I left one more nit. LGTM otherwise I guess.

That's true but I tested all the new targets locally by checking that they were able to do what is expected (updating version, packaging files, ...).

Anyway, I will update the Azure pipeline in different PR(s).

PR #53, opened 2023-08-30
Added missing spotbugs target in Makefile

ppatierno:
Trivial PR to add missing Makefile target for spotbugs which will be needed in the Azure pipeline.

### DIFF
--- a/Makefile.maven
+++ b/Makefile.maven
@@ -27,5 +27,5 @@ java_clean:
 	mvn clean
 
 .PHONY: spotbugs
-spotbugs: java_compile
+spotbugs:
 	mvn $(MVN_ARGS) spotbugs:check
--- a/Makefile.maven
+++ b/Makefile.maven
@@ -25,3 +25,7 @@ java_install:
 java_clean:
 	echo "Cleaning Maven build ..."
 	mvn clean
+
+.PHONY: spotbugs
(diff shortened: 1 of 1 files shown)


scholzj (approved):

Makefile.maven
```diff
@@ -24,4 +24,8 @@ java_install:
 .PHONY: java_clean
 java_clean:
 	echo "Cleaning Maven build ..."
-	mvn clean
\ No newline at end of file
+	mvn clean
+
+.PHONY: spotbugs
+spotbugs: java_compile
```
scholzj, line 30: Do you really want to run `java_compile` before it? That seems like it would cause a lot of time waste in the CI.

ppatierno: But spotbugs needs the new code compiled in order to spot ... bugs :-)
This is what we need when developing locally, or?

scholzj: But it means you will recompile everything in CIs if you do it like this. We do not do that for example in operators.

ppatierno: Agree. It is useful for local development when you could forget to recompile while fixing spotbugs :-) But doesn't make sense in CI. I removed it.

antonio-pedro99 (approved):

kyguy (approved):

PR #55, opened 2023-08-31
Added release and CVE respin pipelines

ppatierno:
This PR adds the release and CVE respin pipelines.

### DIFF
--- a/.azure/release-pipeline.yaml
+++ b/.azure/release-pipeline.yaml
@@ -1,0 +1,56 @@
+# Triggers
+trigger: none
+pr: none
+
+# Parameters
+parameters:
+  - name: releaseVersion
+    displayName: Release Version
+    type: string
+  - name: useSuffix
+    displayName: Build suffixed images
+    type: boolean
+    default: true
+  - name: releaseSuffix
+    displayName: Release Suffix
+    type: number
+  - name: sourcePipelineId
+    displayName: Pipeline ID of the source build
+    type: number
+    default: 59
+    values:
+      - 59
+  - name: sourceBuildId
+    displayName: Build ID of the source build
+    type: number
+
+# Stages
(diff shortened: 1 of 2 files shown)


.azure/release-pipeline.yaml
scholzj, line 1: I guess this would need to do the whole `make release` as well and publish the ZIP / TAR.GZ archives with the installation files and binaries?

ppatierno: I noticed that it was something you had in the access operator but we don't have for the HTTP bridge for example.
When I do the make release locally for the HTTP bridge, I use those generated zip and tar files to be uploaded to the GitHub release.
Are you envisaging to use the ones generated by the release pipeline instead?

scholzj: In that case you are doing it wrong 😉. For Bridge, you should use the ZIPs from Maven. You should not use anything build locally. But as we discussed before - to much value to push the MQTT Bridge to Maven. - so it should be done int he pipeline instead.

ppatierno: I agree with you, in the end I am pushing different zip and tar files than the ones compiled in the pipeline. They are "kind of" the same but not "exactly" the same. Not sure why we didn't do it since the beginning and we don't have it in the HTTP bridge.
I will add this step here and there.

scholzj: I suspect that with TravisCI it was maybe much easier to do it through Maven? That offers the same technical quality. It is just more complicated.

ppatierno: No wait, I was wrong ... For the bridge I am pushing to GitHub release the zip and tar files coming from the target binary so from the actual Azure build. They should be the same pushed to Nexus.

PR #59, opened 2023-11-08
feat: allow the user to set a default Kafka topic for the bridge

### LINKED ISSUE #58: feat: Allow the user to set a default Kafka topic
The default implementation of the default Kafka topic is hard-coded in the`MqttKafkaTopic` interface.
It would be good if the user has the choice to define this topic in the `application.properties` file.

antonio-pedro99:
This PR fixes #58

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,13 +175,13 @@ A valid configuration file should look like this:
 
 The following table describes the configuration properties defined above.
 
-| Setting                        | Description                                                                                      | Default          |
-|--------------------------------|--------------------------------------------------------------------------------------------------|------------------|
-| bridge.id                      | ID of the bridge                                                                                 | my-bridge        |
-| bridge.topic.default | Default Kafka topic for the bridge. If not set, the bridge will use the `messages_default` topic | messages_default |
-| mqtt.host                      | Host address of the MQTT server                                                                  | localhost        |
-| mqtt.port                      | Port number of the MQTT server                                                                   | 1883             |
-| kafka.bootstrap.servers        | Bootstrap servers for Apache Kafka                                                               | localhost:9092   |
+| Setting                 | Description                                              | Default              |
+|-------------------------|----------------------------------------------------------|----------------------|
+| bridge.id               | ID of the bridge                                         | my-bridge            |
+| bridge.topic.default    | Topic to be used if no matches with any mapping rules    | messages_default     |
+| mqtt.host               | Host address of the MQTT server                          | localhost            |
+| mqtt.port               | Port number of the MQTT server                           | 1883                 |
+| kafka.bootstrap.servers | Bootstrap servers for Apache Kafka                       | localhost:9092       |
 
 Other than the above properties, the user can also configure the bridge using environment variables.
 
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ A valid configuration file should look like this:
 ```properties
     # Bridge configuration
     bridge.id=my-bridge
-    bridge.default.topic=default_topic
+    bridge.topic.default=default_topic
     # MQTT configuration
     mqtt.host=localhost
     mqtt.port=1883
@@ -178,7 +178,7 @@ The following table describes the configuration properties defined above.
 | Setting                        | Description                                                                                      | Default          |
 |--------------------------------|--------------------------------------------------------------------------------------------------|------------------|
 | bridge.id                      | ID of the bridge                                                                                 | my-bridge        |
-| bridge.default.topic[Optional] | Default Kafka topic for the bridge. If not set, the bridge will use the `messages_default` topic | messages_default |
+| bridge.topic.default | Default Kafka topic for the bridge. If not set, the bridge will use the `messages_default` topic | messages_default |
 | mqtt.host                      | Host address of the MQTT server                                                                  | localhost        |
 | mqtt.port                      | Port number of the MQTT server                                                                   | 1883             |
 | kafka.bootstrap.servers        | Bootstrap servers for Apache Kafka                                                               | localhost:9092   |
--- a/README.md
+++ b/README.md
@@ -175,13 +175,13 @@ A valid configuration file should look like this:
 
 The following table describes the configuration properties defined above.
 
-| Setting                 | Description                                              | Default              |
-|-------------------------|----------------------------------------------------------|----------------------|
-| bridge.id               | ID of the bridge                                         | my-bridge            |
-| bridge.topic.default    | Topic to be used if no matches with any mapping rules    | messages_default     |
-| mqtt.host               | Host address of the MQTT server                          | localhost            |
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -165,6 +165,7 @@ A valid configuration file should look like this:
 ```properties
     # Bridge configuration
     bridge.id=my-bridge
+    bridge.default.topic=default_topic
```
ppatierno, line 168:
```suggestion
    bridge.topic.default=default_topic
```

I think this should be the other way around so `bridge.topic.default` just in case we would like to set more properties about this topic in the future.

README.md
```diff
@@ -174,12 +175,13 @@ A valid configuration file should look like this:
...
-| Setting                 | Description                        | Default        |
-|-------------------------|------------------------------------|----------------|
-| bridge.id               | ID of the bridge                   | my-bridge      |
-| mqtt.host               | Host address of the MQTT server    | localhost      |
-| mqtt.port               | Port number of the MQTT server     | 1883           |
-| kafka.bootstrap.servers | Bootstrap servers for Apache Kafka | localhost:9092 |
+| Setting                        | Description                                                                                      | Default          |
+|--------------------------------|--------------------------------------------------------------------------------------------------|------------------|
+| bridge.id                      | ID of the bridge                                                                                 | my-bridge        |
+| bridge.default.topic[Optional] | Default Kafka topic for the bridge. If not set, the bridge will use the `messages_default` topic | messages_default |
```
ppatierno, line 181:
```suggestion
| bridge.topic.default | Default Kafka topic for the bridge. If not set, the bridge will use the `messages_default` topic | messages_default |
```

you should explain better that this topic is where messages are forwarded when there is no match with the topic mapping rules.

config/application.properties
```diff
@@ -1,6 +1,6 @@
 #Bridge related settings
 bridge.id=my-bridge
-
+bridge.default.topic=bridge_topic
```
ppatierno, line 3:
```suggestion
bridge.topic.default=default_topic
```

ppatierno (changes requested):

README.md
antonio-pedro99: It is explained early in the docs. However,  I found  it useful to explain here as well. Thanks for the suggestions

README.md
antonio-pedro99: This makes total sense.

ppatierno:
@antonio-pedro99 can you fix the tests failing please?

README.md
```diff
@@ -174,12 +175,13 @@ A valid configuration file should look like this:
...
-| Setting                 | Description                        | Default        |
-|-------------------------|------------------------------------|----------------|
-| bridge.id               | ID of the bridge                   | my-bridge      |
-| mqtt.host               | Host address of the MQTT server    | localhost      |
-| mqtt.port               | Port number of the MQTT server     | 1883           |
-| kafka.bootstrap.servers | Bootstrap servers for Apache Kafka | localhost:9092 |
+| Setting                        | Description                                                                                      | Default          |
+|--------------------------------|--------------------------------------------------------------------------------------------------|------------------|
+| bridge.id                      | ID of the bridge                                                                                 | my-bridge        |
+| bridge.topic.default | Default Kafka topic for the bridge. If not set, the bridge will use the `messages_default` topic | messages_default |
```
ppatierno, line 181: The column separator is not aligned with the rest.
Also it should state something like "Topic to be used if no matches with any mapping rules".
Finally, no need to add "If not set, the bridge will use the `messages_default` topic" because it's clear from the Default column.

src/main/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaMapper.java
```diff
@@ -14,14 +14,15 @@
 public abstract class MqttKafkaMapper {
 
     // default Kafka topic. Used when no mapping rule matches the mqtt topic.
-    public static final String DEFAULT_KAFKA_TOPIC = "messages_default";
+    public static final String MAPPER_DEFAULT_KAFKA_TOPIC = "messages_default";
```
ppatierno, line 17: you are passing the default topic to the mapper so the default should not be here anymore.
It should live in the  `BridgeConfig` class so that it doesn't need a dependency on the mapper.
That is the only place where the default topic constant is used.

ppatierno (changes requested):

kyguy (approved):
Looks good @antonio-pedro99, I don't have any additional comments to what Paolo has already raised, fix those and we  are good to go from my side

README.md
antonio-pedro99: That makes sense. 
In this order, I think we are not handling default bridge ID. We are expecting the user to enter the `.properties` file. 
If the user does not set it, the `bridge-id` will be `null`.

src/main/java/io/strimzi/kafka/bridge/mqtt/mapper/MqttKafkaMapper.java
antonio-pedro99: I agree. It is doable!

antonio-pedro99:
I have added some changes!

README.md
antonio-pedro99: @ppatierno could you take a look at this?

ppatierno: AFAICS the bridge.id is not even actually used so the null value doesn't have any problem.
Also, compared to HTTP bridge we use the bridge.id for generating names to producer and consumer endpoints (something we don't have) and if bridge.id is null we just generate a UUID for those names. But again the bridge.id itself is not used by itself. I think for now we are fine.

README.md
```diff
@@ -175,13 +175,13 @@ A valid configuration file should look like this:
...
-| bridge.id                      | ID of the bridge                                                                                 | my-bridge        |
-| bridge.topic.default | Default Kafka topic for the bridge. If not set, the bridge will use the `messages_default` topic | messages_default |
-| mqtt.host                      | Host address of the MQTT server                                                                  | localhost        |
-| mqtt.port                      | Port number of the MQTT server                                                                   | 1883             |
-| kafka.bootstrap.servers        | Bootstrap servers for Apache Kafka                                                               | localhost:9092   |
+| Setting                 | Description                                              | Default              |
+|-------------------------|----------------------------------------------------------|----------------------|
+| bridge.id               | ID of the bridge                                         | my-bridge            |
```
ppatierno, line 180: as said, we should just say it's null/undefined.

ppatierno (changes requested):

README.md
antonio-pedro99: Yeah, it is fine!

antonio-pedro99:
@ppatierno it should be fine now!

ppatierno (approved):
LGTM!

PR #86, opened 2024-09-24
Implement answer to PingReq MQTT messages (fixes #85)

### LINKED ISSUE #85: Bridge does not answer to PINGREQ
According to the MQTT Spec, a MQTT Server must answer to a PINGREQ message. This bridge does not, which can lead to disconnects from some MQTT Clients, for example rumqttc.

See the MQTT Spec here: http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718085

jhbruhn:
This PR introduces a handler function for PingReq messages, which must be answered with a PingResp message.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
+++ b/src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerHandler.java
@@ -14,6 +14,7 @@ import io.netty.handler.codec.mqtt.MqttPublishMessage;
 import io.netty.handler.codec.mqtt.MqttConnectReturnCode;
 import io.netty.handler.codec.mqtt.MqttConnAckMessage;
 import io.netty.handler.codec.mqtt.MqttQoS;
+import io.netty.handler.codec.mqtt.MqttFixedHeader;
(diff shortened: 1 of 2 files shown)


jhbruhn:
I have added a Unit Test. Unfortunately, paho does not offer a direct way to send a PingReq afaik. The `checkPing` function also just invokes the internal keepAlive mechanism. Thus, I have enabled that mechanism and wrote a test which just checks whether the connection is still intact after 5 seconds with a keepalive interval of 1 second.

ppatierno (approved):
LGTM. Thanks!

kyguy (approved):
To make sure I understand the test correctly, because  of `options.setKeepAliveInterval(1)`  the client in the test sends a PingReq to the server every second, and because the server returns a PingResp back, `client.isConnected()` will be true. Is that the correct understanding?

If so, looks good to me, thanks for the PR @jhbruhn

ppatierno:
Yeah. The keep alive interval is the maximum time between two MQTT messages for the server to understand that the client is alive. In absence of messages (i.e. PUBLISH one), the client needs to keep active traffic by using PINGREQ. I think the test is fine.

PR #125, opened 2026-09-18
Add support for tls/ssl encryption

### LINKED ISSUE #84: No security mechanism implemented
If I am not wrong, there is no authentication/authorization mechanism implemented. Once installed, anyone could send messages to the MQTT endpoint.

Do you know how I could implement at least a simple user and pass authentication? Do you plan to integrate this functionality in the future?

antonio-pedro99:
### Type of change

_Select the type of your PR and delete the other items_

- Enhancement / new feature

### Description

This PR adds support for TLS/SSL Encryption for the MQTT Bridge. It fixes #84 

_Please describe your pull request_

### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [x] Update documentation
- [x] Update CHANGELOG.md (if present)
- [x] Reference relevant issue(s) and close them after merging
- [x] Write tests
- [x] Make sure all tests pass
- [x] Try your changes inside a Kubernetes cluster, not just from unit tests
- [x] AI assistance was used to create this PR (see the [Strimzi AI policy](https://github.com/strimzi/governance/blob/main/AI_POLICY.md))


src/main/java/io/strimzi/kafka/bridge/mqtt/MqttSslContextProvider.java
```diff
@@ -0,0 +1,91 @@
...
+import io.strimzi.kafka.bridge.mqtt.config.MqttSslConfig;
+
+import javax.net.ssl.SSLException;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class MqttSslContextProvider {
```
ppatierno, line 28: Missing Javadoc

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
```diff
@@ -107,11 +108,15 @@ public static void beforeAll() {
 
         // prepare the mapping rules
         String mappingRulesPath = Objects.requireNonNull(MqttBridgetIT.class.getClassLoader().getResource("mapping-rules-regex.json")).getPath();
-        MappingRulesLoader.getInstance().init(mappingRulesPath);
+        try {
+            MappingRulesLoader.getInstance().init(mappingRulesPath);
+        } catch (Exception e) {
+            // no-op, the mapping rules loader is already initialized
+        }
```
ppatierno: why this change?

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttSslIT.java
```diff
@@ -0,0 +1,133 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Random;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
ppatierno, line 36: add a blank line please

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
antonio-pedro99: @ppatierno the MappingRulesLoader is a singleton; it shares state across tests. When I call init in two different tests, the second will throw an exception because the rules have already been initialized. In prod, we initialize the rules once, so I'd say it's okay. Instead of changing the MappingRulesLoader, I am ignoring the exception. But I see one issue: if we just ignore the exception, the mapping rules aren't really being initialized, so we might not even need to call `.init()` in some tests.

README.md
```diff
@@ -151,28 +151,50 @@ Following the prefixes for the specific configurations:
...
+# MQTT SSL/TLS (optional)
+#mqtt.ssl.enable=true
+#mqtt.ssl.certificate.location=path/to/server-cert.pem
+#mqtt.ssl.key.location=path/to/server-key.pem
+# Kafka configuration
+kafka.bootstrap.servers=localhost:9092
+```
 
-The following table describes the configuration properties defined above.
+Uncomment the `mqtt.tls.*` properties to enable TLS/SSL encryption for MQTT connections.
```
kyguy: `mqtt.tls.*` -> `mqtt.ssl.*`

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServer.java
```diff
@@ -46,13 +48,30 @@ public MqttServer(BridgeConfig config, EventLoopGroup masterGroup, EventLoopGrou
...
+     */
+    private MqttSslContextProvider loadSslContextProvider() {
+        if (!mqttConfig.getSslConfig().isEnabled()) {
+            return null;
+        }
+        try {
+            return MqttSslContextProvider.load(mqttConfig.getSslConfig());
+        } catch (SSLException e) {
+            throw new RuntimeException("Failed to load MQTT SSL context", e);
+        }
```
kyguy, line 76: I assume this catch block was added to avoid updating all the connected method signatures with "throws IOExceptions" right?

In a situation where there is an SSL context loading error and we catch and wrap it as a RuntimeException here, is the exception caught by `Main.main()`? Or does the app continue to run as if no error occurred? (I see other exceptions caught in main(),  just not RuntimeException)

As a middle ground, would it make sense to create a new type of exception (e.g. SslContextLoadException) that extends RuntimeException, and wrap with that and then catch that specific exception in "main()" instead?

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
kyguy: Is there a specific exception that is thrown when a multiple rule loaders are initialized? Like a `IllegalStateException`? If so, would it make sense to catch that specific exception instead?

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttSslIT.java
```diff
@@ -0,0 +1,134 @@
...
+    private static MqttConnectOptions tlsConnectOptions() throws Exception {
+        MqttConnectOptions options = new MqttConnectOptions();
+        options.setSocketFactory(SslTestUtils.trustServerCertificateSocketFactory());
+        options.setHttpsHostnameVerificationEnabled(false);
+        options.setConnectionTimeout(10);
+        return options;
+    }
+
+    private String getRandomMqttClientId() {
+        return "mqtt-ssl-client-" + new Random().nextInt(20);
```
kyguy: Would "mqtt-ssl-client-" + UUID.randomUUID() make more sense here as the tests grow?

src/main/java/io/strimzi/kafka/bridge/mqtt/config/MqttSslConfig.java
```diff
@@ -0,0 +1,164 @@
...
+                "config=" + configToString +
+                ")";
+    }
+
+    /**
+     * Hides the inline private key by replacing the actual value with [hidden]
+     *
+     * @return updated configuration with the private key hidden
+     */
+    private Map<String, Object> hideSecrets() {
```
kyguy, line 150: Should we redact `MQTT_SSL_CERTIFICATE` as well?

src/test/java/io/strimzi/kafka/bridge/mqtt/core/MqttServerInitializerTest.java
```diff
@@ -0,0 +1,89 @@
...
+ */
+public class MqttServerInitializerTest {
+
+    @BeforeAll
+    public static void beforeAll() {
+        String mappingRulesPath = Objects.requireNonNull(
+                MqttServerInitializerTest.class.getClassLoader().getResource("mapping-rules-regex.json")).getPath();
+        try {
+            MappingRulesLoader.getInstance().init(mappingRulesPath);
+        } catch (Exception e) {
```
kyguy: Same question here as other comment

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
ppatierno: I think this is not the solution but the symptom of a bad decision we made in the past to make the `MappingRulesLoader` to be singleton. I think we should first fix that. I am working on a POC and let you know when the PR is ready and if/when merged you can rebase this one.

ppatierno: @antonio-pedro99 FYI I opened this PR https://github.com/strimzi/strimzi-mqtt-bridge/pull/126

antonio-pedro99: > I think this is not the solution but the symptom of a bad decision we made in the past to make the MappingRulesLoader to be singleton

Yes, I agree. My first attempt was to refactor it, but I thought it would not be a good idea to change that in this PR, that's why the `hack`. 

I will wait for the PR to be merged; meanwhile, I will leave this comment open

antonio-pedro99: The idea here is to avoid interrupting the tests because of the exception that would have been caught here. 
Please see [this comment](https://github.com/strimzi/strimzi-mqtt-bridge/pull/125#discussion_r4061625740)
@ppatierno has opened a PR to fix the singleton issue, and these `try-catch` blocks will be removed once the PR is merged and I rebase onto this branch.

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttSslIT.java
antonio-pedro99: Yes, it's valid. We can do the same for the MqttIT tests.

src/test/java/io/strimzi/kafka/bridge/mqtt/MqttBridgetIT.java
ppatierno: @antonio-pedro99 the PR is now merged :-)

src/main/java/io/strimzi/kafka/bridge/mqtt/core/MqttServer.java
antonio-pedro99: > I assume this catch block was added to avoid updating all the connected method signatures with "throws IOExceptions" right?

But it would not hurt changing the method signature, there is only one. We could then handle the same exception as we do in `Main.main()`.

kyguy: If it's not too complex let's do that then, I'll defer to your judgement!

PR #126, opened 2026-09-22
Refactored MappingRulesLoader to be an utility class with static load method

ppatierno:
### Type of change

- Refactoring

### Description

While reviewing https://github.com/strimzi/strimzi-mqtt-bridge/pull/125 I realized that the "hack" (catch exception on init) to fix some tests when using the `MappingRulesLoader` is not the cause but the symptom of a useless/bad design of having such class as singleton.
This PR refactors the `MappingRulesLoader` to make it just an utility class with a static method not being singleton anymore.

### Checklist

- [x] Make sure all tests pass
- [x] Try your changes inside a Kubernetes cluster, not just from unit tests
- [x] AI assistance was used to create this PR (see the [Strimzi AI policy](https://github.com/strimzi/governance/blob/main/AI_POLICY.md))


kyguy (approved):
Looks great!

scholzj (approved):

